PR #16, opened 2019-07-02
B64decode

petebryan:
Minor update.
Takes each B64 string, if its not divisible by 4 and therefore not a fully formatted B64 string pads out the string with 'A's. This helps if string selected from log files are truncated for any reason.


petebryan:
Formatting errors being generated from existing library files, I've seen a lot of the formatting changes have been resolved in the TI Branch Ian is working from. I will wait for that to commit and then rebase rather than duplicate effort

petebryan:
Merged with master, should be good to go now

PR #17, opened 2019-07-16
Geo lite lookup dbupdate

ashwin-patil:
Changelog :: 
Updates to geolitelookup class
- Use official maxmind pypi package - geoip2
- download maxmind city database locally in msticpy package directory or user specified.
-  ability to specify existing path containing local maxmind city database.
-  ability to issue force update
-  Check age of maxmind city database based on database info and download new if it is not updated in last 30 days.
update geoip.py file with black formatting


msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+                    print(f"Extracting city database...")
+                    with open(_DB_FILE_PATH, "wb") as f_out:
+                        shutil.copyfileobj(f_in, f_out)
+                        print(
+                            f"Extraction complete. Local Maxmind city DB :: {_DB_FILE_PATH}"
+                        )
+            except IOError as err:
+                print(f"{_DB_ARCHIVE_PATH} {err}")
+
+    def get_geolite_dbpath(
```
ianhelle: Split into 
1. get_geolite_dbpath property method (@property decorator) that returns the current path
2. _check_and_update_version (or some appropriate name)

msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+            print(
+                f"No local Maxmind City Database found. Downloading new database to {db_folder}"
+            )
+            self.download_and_extract_gzip(self._MAXMIND_DOWNLOAD, db_folder)
+            _list_of_db_paths = glob.glob(db_folder + "/*.mmdb")
+            _latest_db_path = _list_of_db_paths[0]
+
+        return _latest_db_path
+
+    def __init__(self, db_folder: str = None, force_update: bool = False):
```
ianhelle: Keep as top method in class.
Add "auto_update" parameter to allow people to disable auto-update

msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+                self.download_and_extract_gzip(self._MAXMIND_DOWNLOAD, db_folder)
+            else:
+                if force_update:
+                    print(
+                        f"Force update is set to True. Downloading new database to {db_folder}"
+                    )
+                    self.download_and_extract_gzip(self._MAXMIND_DOWNLOAD, db_folder)
+                else:
+                    print(
+                        "Existing local Maxmind City Database is downloaded in last 30 days so it will be used"
```
ianhelle: Not sure you need this - only send messages when the user needs to do something (unless you want a verbose option)

msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+        Returns
+        -------
+        str
+            Returns the absoulte path of local maxmind geolite city database after control flow logic.
+
+        """
+        if db_folder is None:
+            db_folder = self._PKG_DIR
+        force_update = self._force_update
+        _list_of_db_paths = glob.glob(db_folder + "/*.mmdb")
```
ianhelle: You don't need underscores for local vars - since these are always private to the function

msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+            Provide absolute path to the folder containing MMDB file (e.g. '/usr/home' or 'C:\maxmind'). 
+            If no path provided, it is set to download msticpy package installation directory.
+        force_update : bool, optional
+            Force update can be set to true or false. depending on it, new download request will be initiated.
+
+        """
+        if db_folder is None:
+            db_folder = self._PKG_DIR
+        self._force_update = force_update
+        self._dbpath = self.get_geolite_dbpath(db_folder, self._force_update)
```
ianhelle: Make separate calls to get dbpath and check & update.
If there is no dbpath found or auto_update (True) you can go to check_and_update, which should return the latest_dbpath.
Surround call to check_and_update with try/except.
- if it threw an exception AND there was a previous dbpath - you can issue a warning (import warnings, warnings.warn(...) and just use the previous dbpath
- if an exception and no previous dbpath you should exit with an error - probably rethrow the exception you caught, since this will have the most relevant info about what went wrong. (just use "raise" with no args inside the except block - this will rethrow the same exception and keep the call stack intact.

msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+
+        Returns
+        -------
+        str
+            Returns the absoulte path of local maxmind geolite city database after control flow logic.
+
+        """
+        if db_folder is None:
+            db_folder = self._PKG_DIR
+        force_update = self._force_update
```
ianhelle: This is overwriting the value of the parameter. Can just delete this line since it is sent from __init__ correctly

msticpy/sectools/geoip.py
```diff
@@ -278,27 +293,139 @@ class GeoLiteLookup(GeoIpLookup):
...
+        _list_of_db_paths = glob.glob(db_folder + "/*.mmdb")
+
+        if _list_of_db_paths:
+            _latest_db_path = max(_list_of_db_paths, key=os.path.getmtime)
+            print(f"Local Maxmind City Database already present in {db_folder}")
```
ianhelle: Remove this print() unless you want a verbose option

ashwin-patil: removed

msticpy/sectools/geoip.py
ashwin-patil: deleted the assignment

msticpy/sectools/geoip.py
ashwin-patil: added option to toggle auto_update and separate call to check_and_update function and get_geoip_db_path to return latest_dbpath

msticpy/sectools/geoip.py
ashwin-patil: renamed the local vars without underscores

msticpy/sectools/geoip.py
ashwin-patil: agreed. removed additional print messages.

msticpy/sectools/geoip.py
ashwin-patil: added option to toggle auto_update

msticpy/sectools/geoip.py
ashwin-patil: need to check about decorator in the current modified code. have added new function _check_and_update.

PR #22, opened 2019-08-08
Fix the broken links to notebooks on readthedocs

weichea:
Fix the broken links.


ianhelle:
@weichea - thanks for the fixes. I wasn't able to get the pipeline to build this branch so I've put your fixes and a couple of others in this commit #25 . These should get into master soon.

PR #23, opened 2019-08-20
TI Providers - Az Sentinel BYOTI

ianhelle:
Added provider for Azure Sentinel byoti (plus base class for Kusto-based TI sources)
Unit tests for this and re-wrote unit test for http providers.
Fixed broken lookup_iocs() method for http providers.


petebryan:
ti_provider_base - preprocess_observable.

It checks whether the observable matches the format for the IOC type specified, but it doesn't check the value provided for IOC type at all. So if I provide an IOC type of 'rubbish' it treats it fails the validation step as observable not matching type. It might be nice to do some sort of validation of IOC type as well and return if IOC Type is not supported to make it clearer what the issue is.

ianhelle:
@petebryan - fixed the doc bug and the check_ioc_type error.
Also some issues with reading config and kql_driver reporting 0 row results as an error

PR #24, opened 2019-08-20
Fix nbwidgets Lookback._time_range_change

anoryx:
Change of Lookback slider sets "start" time in past, not future.


ianhelle:
@anoryx - thanks for the fix. Sorry it took so long to get in. We found that external contribs were not getting built by Azure pipelines. Fixed now and merged

PR #26, opened 2019-09-07
Multi timeline plot

ianhelle:
Merged multi-timeline plots into single API
Implemented pre-commit hook to update TLD seed file.
Fixed TILookup problems to alert when not properly configured.


msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+        x_axis_type="datetime",
+        x_minor_ticks=10,
+        y_axis_label=y,
+        tools=[hover, "xwheel_zoom", "box_zoom", "reset", "save", "xpan"],
+        toolbar_location="above",
+        title=title,
+    )
+    plot.yaxis.visible = show_yaxis
+    plot.ygrid.minor_grid_line_color = "navy"
+    plot.ygrid.minor_grid_line_alpha = 0.3
```
petebryan: These minor lines don't really add anything except clutter the graph, suggest we set minor_grid_line_color = None

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+    plot.xgrid.minor_grid_line_color = "navy"
+    plot.xgrid.minor_grid_line_alpha = 0.3
```
petebryan: Could also potentially drop the x minor lines as well to make the timeline cleaner.

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+        Add a reference line/label using the alert time (the default is None)
+    ref_time : datetime, optional
+        Add a reference line/label using `ref_time` (the default is None)
+    group_by : str
+        (where `data` is a DataFrame)
+        The column to group timelines on
+    sort_by : str
+        (where `data` is a DataFrame)
+        The column to order timelines on
+    legend: str, optional
```
petebryan: I think the legend is pretty key to comprehension of the timeline, I would suggest we default the legend to being inline and provide the option to remove/move it if wanted.

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+        left, right or inline
+        (the default is None/no legend)
+    legend_column : str, optional
+        (where `data` is a DataFrame)
+        Name of the column used to generate the legend labels if a legend is
+        to be displayed. Default is `group_by` parameter.
+    yaxis : bool, optional
```
petebryan: Is there ever a need to display the yaxis labels given we are just assigning arbitrary values for this?

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+        plot.yaxis.visible = True
+        if data:
+            y_labels = {ser_def["y_index"]: str(lbl) for lbl, ser_def in data.items()}
+            plot.yaxis.major_label_overrides = y_labels
+    plot.ygrid.minor_grid_line_color = "navy"
+    plot.ygrid.minor_grid_line_alpha = 0.3
+    plot.xgrid.minor_grid_line_color = "navy"
+    plot.xgrid.minor_grid_line_alpha = 0.3
+    # Create plot bar to act as as range selector
+    rng_select = _create_range_tool(
```
petebryan: We make this optional for the other graphs, we should probably do the same here.

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+    """Add a reference marker line and label at `ref_time`."""
+    ref_label_tm = pd.Timestamp(ref_time)
+    plot.line(x=[ref_label_tm, ref_label_tm], y=[0, series_count])
+    ref_label = Label(
+        x=ref_label_tm,
+        y=0,
+        y_offset=10,
+        x_units="data",
+        y_units="data",
+        text="< ref",
```
petebryan: This text isn't particularly useful when using the timeline, we should have it as a kwarg we can pass with the ref object or if we pass it an alert at least set this to 'Alert'

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+def _create_range_tool(
+    data, min_time, max_time, plot_range, width, time_column: str = None
+):
+    """Create plot bar to act as as range selector."""
+    ext_min = min_time - ((max_time - min_time) * 0.15)
+    ext_max = max_time + ((max_time - min_time) * 0.15)
+    rng_select = figure(
+        x_range=(ext_min, ext_max),
+        title="Range Selector",
+        plot_height=100,
```
petebryan: This is too small on a high res display.  Maybe if we set this height as a percentage of the the main timeline height.

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+    # build the tool tips from all specified columns
+    tool_tip_items = [(f"{col}", f"@{col}") for col in tool_tip_columns]
+    hover = HoverTool(tooltips=tool_tip_items, formatters={"Tooltip": "printf"})
+
+    title = f"Timeline: {title}" if title else "Event Timeline"
+    start_range = min_time - ((max_time - min_time) * 0.1)
+    end_range = max_time + ((max_time - min_time) * 0.1)
+    height = height if height else _calc_auto_plot_height(len(data))
+    plot = figure(
+        x_range=(start_range, end_range),
```
petebryan: for the multi data set plots we probably want to set the y_range as well. Something like y_range=(0, len(data))

msticpy/nbtools/timeline.py
```diff
@@ -0,0 +1,940 @@
...
+
+    title = f"Timeline: {title}" if title else "Event Timeline"
+    min_time = graph_df[time_column].min()
+    max_time = graph_df[time_column].max()
+    start_range = min_time - ((max_time - min_time) * 0.1)
+    end_range = max_time + ((max_time - min_time) * 0.1)
+    height = height if height else _calc_auto_plot_height(series_count)
+
+    plot = figure(
+        x_range=(start_range, end_range),
```
petebryan: We probably also want to set y_range here.

msticpy/nbtools/timeline.py
ianhelle: Will make yaxis minor lines fainter, make xaxis on by default but suppressable.

msticpy/nbtools/timeline.py
ianhelle: Legend on (inline, centered vertically) by default unless single data series with no legend column

msticpy/nbtools/timeline.py
ianhelle: make yaxis always optional

msticpy/nbtools/timeline.py
ianhelle: implement reference label param (infer if alert supplied)

msticpy/nbtools/timeline.py
ianhelle: Make 20% of main chart size

msticpy/nbtools/timeline.py
ianhelle: Agreed - +/- 25% of min/max

msticpy/nbtools/timeline.py
ianhelle: Agreed - +/- 25% of min/max

PR #31, opened 2019-10-18
Added domain utilities

petebryan:
Added:
- Domain utils package for validating a domain name, and taking screenhots via Browshot
   - Relocated some code from iocextract to avoid duplication
- Tests for package

Fixed:
- Query error where column was renamed but original name was then later referenced
- Map center calculation now handles if no locations are passed to it rather than just failing
- Minor formatting

### DIFF
--- a/msticpy/sectools/domain_utils.py
+++ b/msticpy/sectools/domain_utils.py
@@ -16,12 +16,12 @@ import ssl
 import json
 import time
 import requests
-import tldextract
 import dns.resolver
 import OpenSSL
 import pandas as pd
 from ipywidgets import IntProgress
 from IPython import display
+import tldextract
 from ..nbtools import pkg_config as config
 from .._version import VERSION
 from ..nbtools.utility import export
--- a/msticpy/sectools/domain_utils.py
+++ b/msticpy/sectools/domain_utils.py
@@ -68,7 +68,7 @@ def screenshot(url: str, api_key: str = None) -> requests.models.Response:
     status_string = (
         f"https://api.browshot.com/api/v1/screenshot/info?id={bs_id}&key={bs_api_key}"
     )
-    image_string = f"https://api.browshot.com/api/v1/screenshot/thumbnail?id={bs_id}&zoom=50&key={bs_api_key}"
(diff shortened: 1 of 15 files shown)


tests/test_tiproviders.py
```diff
@@ -404,11 +404,11 @@ def test_tor_exit_nodes(self):
             else:
                 neg_results.append(lu_result)
 
-        self.assertEqual(len(pos_results), 4)
```
ianhelle: I think I fixed this in iputils - since this list is going to be changing I'm taking the known Tor nodes as a sample from the current list and testing against that - so unless one of the neg-results set becomes a Tor node the test should continue to work. We can deal with this when we merge though.

tests/test_domain_utils.py
```diff
@@ -0,0 +1,18 @@
...
+# license information.
+# --------------------------------------------------------------------------
+"""domain_utilstes extract test class."""
+
+from ..msticpy.sectools import domain_utils
+
+
+def test_validate_domain():
+    test_dom_val = domain_utils.DomainValidator()
+    valid_tld = test_dom_val.validate_tld("www.microsoft.com")
```
ianhelle: I know we don't have many negative tests but is it worth putting some known broken domain to act as a negative, given the unit test is pretty simple?

msticpy/sectools/domain_utils.py
```diff
@@ -0,0 +1,254 @@
...
+    status_string = (
+        f"https://api.browshot.com/api/v1/screenshot/info?id={bs_id}&key={bs_api_key}"
+    )
+    image_string = f"https://api.browshot.com/api/v1/screenshot/thumbnail?id={bs_id}&zoom=50&key={bs_api_key}"  # pylint: disable=line-too-long
+    # Wait until the screenshot is ready and keep user updated with progress
+    print("Getting screenshot")
+    progress = IntProgress(min=0, max=40)
+    display.display(progress)
+    ready = False
+    while ready is False:
```
ianhelle: Nice. I should do something like this in the TI provider lookup

tests/test_domain_utils.py
petebryan: Good idea, I will add one.

PR #34, opened 2019-10-29
Iputils

ashwin-patil:
Changelog:

- Added iputils package with whois, and IP related functions.
- Added new yaml queries for Heartbeat, timeseries multidatasources.
- added utility to check and install missing packages for Setup section of notebook.
- updated requirements with new dependencies - ipwhois, tqdm
- fixed black formatting and pylint warnings for multiple files based on build warnings.


msticpy/sectools/ip_utils.py
```diff
@@ -0,0 +1,200 @@
...
+    """
+    ip_type = get_ip_type(ip_str)
+    if ip_type == "Public":
+        whois = IPWhois(ip_str)
+        whois_result = whois.lookup_whois()
+        if show_progress:
+            print(".", end="")
+        return whois_result["asn_description"], whois_result
+    else:
+        return "NO ASN Information since IP address is of type", {ip_type}
```
ianhelle: This return looks a bit weird. Did you mean to append the ip_type to the error message and return an empty dict {}?

msticpy/sectools/ip_utils.py
```diff
@@ -0,0 +1,200 @@
...
+        elif " " in ip_str:
+            addrs = ip_str.split(" ")
+        else:
+            addrs = [ip_str]
+
+        for addr in addrs:
+            ip_entity = IpAddress()
+            ip_entity.Address = addr.strip()
+            try:
+                IPLOCATION.lookup_ip(ip_entity=ip_entity)
```
ianhelle: Does this throw a DataError?

msticpy/sectools/ip_utils.py
```diff
@@ -0,0 +1,200 @@
...
+from ..nbtools.utility import export
+from .geoip import GeoLiteLookup
+
+__version__ = VERSION
+__author__ = "Ashwin Patil"
+
+IPLOCATION = GeoLiteLookup()
+
+
+class Error(Exception):
```
ianhelle: I think it's good to have custom exceptions - I regret not doing that from the beginning. It would be good to have maybe more descriptive names though if you are going to create custom exceptions. 
Also I don't see that you've used them anywhere?

petebryan: Think this might be a carry over from syslog_utils where I have the DataErrror custom exception.

msticpy/sectools/ip_utils.py
ashwin-patil: f-string typo, should be part of string with ip_type value replaced. I will fix this.

msticpy/sectools/ip_utils.py
ashwin-patil: yup, carried over from syslog_utils. I will check and remove if these are unused in ip_utils.

PR #35, opened 2019-11-25
Bug/timeline

petebryan:
Bokeh 1.4.0 changes how legends are defined:
https://docs.bokeh.org/en/latest/docs/releases.html

Specifically: 
"""
The overburdened legend keyword argument to glyph methods is deprecated. It is replaced with three purpose-specific keyword arguments:

legend_label="some_label"

This will always produce a legend item with exactly the given label.

legend_field="some_colname"

This will produce a “grouped” legend, where the grouping is done by JavaScript in the browser. Python code will only see a single legend item representing the grouping.

legend_group="some_colname"

This will produce a “grouped” legend, where the grouping is done by Python code, bedore the content is rendered in the browser. Python code will see individual legend items for each group.
"""

This PR updates the timeline function to use the new legend keyword arguments.
It also addresses updates the logic in places to address some issues found when None is passed as the legend_label argument.

### DIFF
--- a/msticpy/nbtools/timeline.py
+++ b/msticpy/nbtools/timeline.py
@@ -295,7 +295,10 @@ def display_timeline_values(
             p_series = []
             # create default plot args
             plot_args: Dict[str, Any] = dict(
-                x=time_column, alpha=0.7, source=row_source, legend_label=str(inline_legend)
+                x=time_column,
+                alpha=0.7,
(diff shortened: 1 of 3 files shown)


ianhelle:
We should probably update the requirements/setup to Bokeh 1.4 if this change breaks earlier versions.

PR #38, opened 2019-12-14
Process tree 2

ianhelle:


msticpy/nbtools/process_tree.py
```diff
@@ -0,0 +1,430 @@
...
+    rng_select = figure(
+        plot_width=width, plot_height=height, y_range=(min_y - 1, max_y + 1)
+    )
+
+    x_dodge = dodge(x_col, -0.5)
+    rng_select.rect(
+        x=x_dodge,
+        y=y_col,
+        width=1.2,
+        height=0.8,
```
petebryan: I am finding this too small to be usable if you have more than a small set of data - suggest upping it to 1.2

msticpy/sectools/process_tree_utils.py
```diff
@@ -0,0 +1,881 @@
...
+    return proc_tree
+
+
+def infer_schema(data: Union[pd.DataFrame, pd.Series]) -> ProcSchema:
+    """
+    Infer the correct schema to use for this data set.
+
+    Parameters
+    ----------
+    procs : pd.DataFrame
```
petebryan: variable name should be data

msticpy/data/queries/kql_sent_lxauditd.yaml
```diff
@@ -0,0 +1,56 @@
...
+    description: Extract all audit messages grouped by mssg_id
+    args:
+      query: '
+        {table}
+        | where {subscription_filter}
+        | where TimeGenerated >= datetime({start})
+        | where TimeGenerated <= datetime({end})
+        | where Computer has "{host_name}"
+        // extract the message parts from the RawData field
+        | extend mssg_parts = extract_all(@"type=(?P<type>[^\s]+)\s+msg=audit\((?P<mssg_id>[^)]+)\):\s+(?P<mssg>[^\r]+)\r?",
```
petebryan: Comments don't work in these queries - its concats the query to one string so anything after the first comment line gets commented out.

msticpy/data/queries/kql_sent_lxauditd.yaml
```diff
@@ -0,0 +1,56 @@
...
+          dynamic([\"type\", \"mssg_id\", \"mssg\"]), RawData)
```
petebryan: don't need the \ here (it causes query errors)
 dynamic(["type", "mssg_id", "mssg"]), RawData) - works fine

msticpy/nbtools/process_tree.py
ianhelle: Pretty sure this is the glyph height so don't see how this affects the range box height

msticpy/data/queries/kql_sent_lxauditd.yaml
ianhelle: Looks like you fixed this.

msticpy/data/queries/kql_sent_lxauditd.yaml
ianhelle: and this

PR #48, opened 2020-02-21
Azure data expansion

petebryan:
New Features:
- Updates to Azure Data package to allow for querying of additional data sources including network interface details and resource metrics.
- Improved test coverage

Fixes:
- OData driver fixes to allow connection at initialization and connection string usage
-  Fixed timeline bar color selection
- Provided support for get_map_center with current Notebooks
- Expanded use of MsticpyException
- Reduced Azure package requirements to allow usage with other packages
- Added clarification to documentation
- Fixed azure-mgmt package renaming in import_analyzer


msticpy/nbtools/timeline.py
```diff
@@ -250,12 +250,12 @@ def display_timeline_values(
 
     reset_output()
     output_notebook()
-
     height: int = kwargs.pop("height", None)
     width: int = kwargs.pop("width", 900)
     title: str = kwargs.pop("title", None)
     time_column = kwargs.get("x", time_column)
     group_by: str = kwargs.get("group_by", None)
+    legend_column: str = kwargs.pop("legend_column", group_by)
```
ianhelle: I think I removed legend_column from the other timelines since it duplicated group_by and didn't actually work (as far as I remember). I'm thinking if "Account" is specified as group_by and "Computer" is specified as legend_column - what happens?

msticpy/nbtools/timeline.py
```diff
@@ -320,18 +320,15 @@ def display_timeline_values(
...
             plot_args: Dict[str, Any] = dict(
-                x=time_column,
-                alpha=0.7,
-                source=row_source,
-                legend_label=str(inline_legend),
+                x=time_column, alpha=0.7, source=row_source, legend_group=legend_column
             )
             if "vbar" in plot_kinds:
-                p_series.append(plot.vbar(top=y, width=4, color="color", **plot_args))
+                p_series.append(plot.vbar(top=y, width=4, color=color, **plot_args))
```
ianhelle: Doesn't this set all series to a single color? The "color" parameter here is meant to reference the value set in the call to _create_data_grouping which is mapped for each group. Maybe we should re-think this - how a user can specify the color map to use or something.

msticpy/nbtools/timeline.py
```diff
@@ -868,10 +865,16 @@ def _create_range_tool(
 # pylint: enable=too-many-arguments
 
 
-def _add_ref_line(plot, ref_time, ref_text="Ref time", series_count=1):
+def _add_ref_line(plot, ref_time, ref_text="Ref time", series_count=100):
```
ianhelle: Not sure why the default of 100 is needed. I think I only put a default there because the ref_text param has a default and you can't have a subsequent param without a default. I think in each case we call this, we specify the value.

msticpy/data/azure_data.py
```diff
@@ -298,14 +369,13 @@ def get_api(
...
 
-        if self.resource_client is None:
-            self.resource_client = ResourceManagementClient(self.credentials, sub_id)
-            if not self.resource_client:
-                raise CloudError("Could not create a ResourceManagementClient.")
+        self._check_client("resource_client", sub_id)  # type: ignore
 
+        # Normalise elements depending on user input type
         if resource_id is not None:
             namespace = resource_id.split("/")[6]
```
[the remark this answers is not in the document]
petebryan: wrapped in try/except with an example of expected format in exception.

msticpy/data/azure_data.py
```diff
@@ -333,3 +406,207 @@ def get_api(
...
+        Parameters
+        ----------
+        metrics: str
+            A string list of metrics you wish to collect
+            (https://docs.microsoft.com/en-us/azure/azure-monitor/platform/metrics-supported)
+        resource_id: str
+            The resource ID of the resource to collet the metrics from
+        sub_id: str
+            The subscription ID that the resource is part of
+        sample_time: str (Optional)
```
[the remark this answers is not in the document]
petebryan: Done

msticpy/data/azure_data.py
```diff
@@ -333,3 +406,207 @@ def get_api(
...
+        start_time: int (Optional)
+            The number of days prior to today to collect metrics for, default is 30
+
+        Returns
+        -------
+        results: dict
+            A Dictionary of DataFrames containing the metrics details
+
+        """
+        if sample_time.lower() == "h" or sample_time.lower() == "hour":
```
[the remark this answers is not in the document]
petebryan: Done and updated to casefold - thanks.

msticpy/nbtools/timeline.py
petebryan: Accidently left in from testing to try and solve why a legend was showing even then legend='none' was set. Fixed now with working code.

msticpy/nbtools/timeline.py
petebryan: Yep - changed.

msticpy/nbtools/timeline.py
petebryan: Had some feedback from user saying that the line 'never showed'. Turned out it was showing but because it was len of group_by data set rather than related to the y_axis value size it was creating a very small line instead. This change was part of testing what was happening. Have included the actual fix now which takes the max y_axis value and uses that as the series_count size to pass to _add_ref_line.

PR #55, opened 2020-04-03
Safe String queries

petebryan:
OTX doesn't support URLs with unsafe characters in.
This update adds a property to TI Providers of require_url_encoding which if set to True will encode the URL before passing to the provider.

In addition if a provider does not return properly formatted JSON data an error will be displayed in the results rather than an exception being thrown to ensure one bad result doesn't prevent other responses being obtained.

### DIFF
--- a/docs/source/getting_started/JupyterAndAzureSentinel.rst
+++ b/docs/source/getting_started/JupyterAndAzureSentinel.rst
@@ -179,10 +179,6 @@ References
    is the official documentation for using Jupyter notebooks in Azure
    Sentinel.
 
--  `Get Started <https://github.com/Azure/Azure-Sentinel-Notebooks/blob/master/Get%20Started.ipynb>`__
-   is a Jupyter notebook that demonstrates the use of Kqlmagic to
-   retrieve data from Azure Sentinel.
-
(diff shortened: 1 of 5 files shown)


msticpy/sectools/tiproviders/ti_provider_base.py
```diff
@@ -463,9 +473,11 @@ def _check_ioc_type(
 
 
 # pylint: disable=too-many-return-statements, too-many-branches
-def preprocess_observable(observable, ioc_type) -> SanitizedObservable:
+def preprocess_observable(
+    observable, ioc_type, require_url_encoding: bool = False
+) -> SanitizedObservable:
     """
-    Preprocesses and checks validity of observable against declared IoC type.
+    Preprocesses and checks validity observable against declared IoC type.
```
ianhelle, line 480: why delete "of"?

petebryan: Honestly I have no idea. I have re-instated the of.

petebryan: Honestly have no idea, I've re-instated.

PR #58, opened 2020-04-18
Morph charts

petebryan:
New Features:
- Morph Chart Module for formatting and displaying morph charts in a Notebook.
- Templated queries and chart descriptions for charts based on Signin Logs.
- Associated tests and documentation.
- Added option to return figure rather than displaying when calling display_timeline.

Fixes:
- Removed redundent code from test_azuredata

Notes:
Code coverage for new features - 96.4%


msticpy/data/queries/kql_sent_azure.yaml
```diff
@@ -123,4 +123,20 @@ sources:
       resource:
         description: ResourceId or substring
         type: str
+  morph_signins:
```
ianhelle, line 126: Seems like an odd name for a pretty generic query - maybe AADSigninsWithGeogLoc ?

msticpy/nbtools/morph_charts.py
```diff
@@ -0,0 +1,156 @@
...
+
+        chart_name: str:
+            The name of the Morph Chart to plot.
+
+        """
+        # Check input data is correct format and that the chart being requested exists
+        if not isinstance(data, pd.DataFrame):
+            raise MsticpyException("Data provided must be in pandas.DataFrame format")
+
+        try:
```
ianhelle, line 46: You could just use something like self.charts.get(chart_name) or if chart_name not in self.charts:

msticpy/nbtools/morph_charts.py
```diff
@@ -0,0 +1,156 @@
...
+        # Display Morph Charts in IFrame with instructions
+        print(
+            f"Navigate to {Path.cwd().joinpath('morphchart_package')} and upload the files below"
+        )
+        print("Charts provided by http://morphcharts.com/")
+        return IFrame("http://morphcharts.com/designer.html", "100%", "600px")
+
+    def list_charts(self):
+        """Get a list of avaliable charts."""
+        for key, _ in self.charts.items():
```
ianhelle, line 76: more simply "for key in self.charts"

msticpy/nbtools/morph_charts.py
```diff
@@ -0,0 +1,156 @@
...
+                print(key, ":", "\n", self.charts[key]["Description"])
+            elif keyword.casefold() in [
+                word.casefold() for word in value["Description"].split()
+            ]:
+                print(key, ":", "\n", self.charts[key]["Description"])
+            else:
+                print("No matching charts found")
+
+
+def _get_charts(path: str = "MorphCharts") -> dict:
```
ianhelle, line 123: lower case folder name?

msticpy/nbtools/morph_charts.py
```diff
@@ -0,0 +1,156 @@
...
+    full_path = Path(__file__).parent.parent.joinpath("data").joinpath(path)
+    file_glob = Path(full_path).glob(f"*.yaml")
+    chart_files = [file_path for file_path in file_glob if file_path.is_file()]
+    chart_details = {}
+    for chart in chart_files:
+        chart_data = open(chart, "r")
+        details = yaml.safe_load(chart_data)
+        chart_details.update(
+            {
+                details["Name"]: {
```
ianhelle, line 147: Might want to give more feedback about input given that we might not control the file. I guess the dict key failure would would tell people that they didn't have the correct file format. Perhaps just wrap this in a try/except and throw an exception saying that the file is invalid format.

PR #62, opened 2020-05-04
Pkg config, nbinit and other fixes

ianhelle:
This includes a few changes to the package config stuff:
- Validation function for msticpy config
- Enable Key Vault secrets for Azure CLI and Azure Sentinel
- Wsconfig has better validation, error messages and logic to always fall back to config.json if msticpyaml doesn't contain the right data.
NBinit.py contains a function that encapsulates a lot of the Setup cell.
Updates to install_missing packages to honor package versions
Linux logon failure queries (for Password spray notebook)
Timeseries and Timeline plots now return whole layout (including range tool) rather than just the main plot
Iputils - added populate_host_entity (same as populate_ip_entity) but works with a host_entity input


msticpy/common/pkg_config.py
```diff
@@ -25,14 +25,17 @@
...
 from .._version import VERSION
 
 __version__ = VERSION
 __author__ = "Ian Hellen"
 
 _CONFIG_FILE: str = "msticpyconfig.yaml"
 _CONFIG_ENV_VAR: str = "MSTICPYCONFIG"
+_DP_KEY = "DataProviders"
+_AZ_SENTINEL = "AzureSentinel"
+_AZ_CLI = "AzureCLI"
```
petebryan, line 38: If we are hardcoding this should we update our msticpy.rst file example to include this?

msticpy/common/pkg_config.py
```diff
@@ -200,5 +204,178 @@ def _get_top_module():
...
+    return mp_errors, mp_warnings
+
+
+def _check_provider_settings(mp_config, section, key_provs):
+    mp_errors = []
+    mp_warnings = []
+    if not mp_config:
+        mp_warnings.append(f"'{section}' section has no settings.")
+    for p_name, p_setting in mp_config.items():
+        sec_args = p_setting.get("Args")
```
petebryan, line 329: Throws an attribute error if you call validate_config() with config_file set. 
If config_file contains AzureCli and has an empty QueryDefinitions section this gets called with with ('QueryDefinitions', None) and tries get("Args")  on None (my config file was like this).
Could catch where p_setting is none and raise this as an mp_warning

msticpy/data/queries/kql_sent_lxsyslog_logon.yaml
```diff
@@ -4,6 +4,34 @@ metadata:
   data_environments: [LogAnalytics]
   data_families: [LinuxSyslog]
   tags: ['linux', 'syslog']
+  aliases:
```
petebryan, line 7: We should probably document this feature (couldn't find anything referencing aliases)

msticpy/common/pkg_config.py
ianhelle: Done. I;ve documented the AzureCLI as going in the DataProviders section. It can actually go in either. I;ve updated the config validation to account for it being in either place.

msticpy/data/queries/kql_sent_lxsyslog_logon.yaml
ianhelle: It's apparently a standard yaml feature. Maybe we should document this whole query template stuff a little better.

msticpy/common/pkg_config.py
ianhelle: I think I've fixed this - it should have only been sending the AzureCLI section. I've pulled it out into a temp dictionary so that it matches the pattern of the TIProviders (i.e. how far down the dict tree the provider settings are).
If AzureCLI settings are in the DataProviders section this will have been processed already

PR #69, opened 2020-06-08
Statsmodels timeseries

ashwin-patil:
Additions/Changes:
- added time series anomalies function using STL method
- Added test case for the function
- Updated documentation notebook TimeSeriesAnomaliesVisualization with new function
- ReadTheDocs page update


setup.py
```diff
@@ -46,8 +46,10 @@
     "pyyaml>=3.13",
     "requests>=2.21.1",
     "scikit-learn>=0.20.2",
+    "scipy>=1.4.1",
```
ianhelle, line 49: Are you able to work with scipy>=1.1.0 and statsmodels>=0.9.0?
These are the versions already in Azure Notebooks and would save us some install time if we didn't have to upgrade. AzureML should be ok - it has later versions of these.

requirements.txt
```diff
@@ -34,9 +34,12 @@ pytz>=2019.2
 pyyaml>=3.13
 requests>=2.21.1
 scikit-learn>=0.20.2
+scipy>=1.4.1
 seaborn>=0.9.0
 setuptools>=40.6.3
+statsmodels>=0.11.1
```
ianhelle, line 40: Can you add these the conda/conda-reqs.txt as well (otherwise you'll get a test failiure)

docs/source/visualization/TimeSeriesAnomalies.rst
```diff
@@ -345,6 +343,299 @@ the similar details
     </table>
     </div>
 
+Using MSTICPY - Seasonal-Trend decomposition using LOESS (STL)
+--------------------------------------------------------------
+
+In this case, we will use function msticpy function `timeseries_anomalies_stl` which leverages `STL` method from `statsmodels` API to decompose a time series into three components: trend, seasonal and residual. STL uses LOESS (locally estimated scatterplot smoothing) to extract smooths estimates of the three components. The key inputs into STL are:
```
petebryan, line 349: "we will use function msticpy function `timeseries_anomalies_stl` "
Think this should be "we will use the msticpy function `timeseries_anomalies_stl` "

msticpy/nbtools/timeseries.py
```diff
@@ -242,3 +244,76 @@ def display_timeseries_anomolies(
...
+        by default 3
+
+    Returns
+    -------
+    pd.DataFrame
+        Returns a dataframe with additional columns by decomposing time series data
+        into residual, trend, seasonal, weights, baseline, score and anomalies.
+        The anomalies column will have 0, 1,-1 values based on score_threshold set.
+    """
+    seasonal: int = kwargs.get("seasonal", 7)
```
ianhelle, line 278: You might want to use the common.utils.check_kwargs() function here. It checks for bad arguments and will suggest closest matches (e.g. if someone sends threshold=20 it will tell them that the closest match is score_threshold)

msticpy/nbtools/timeseries.py
```diff
@@ -8,6 +8,8 @@
 from itertools import zip_longest
```
ianhelle, line 8: Can you extract this into a separate module and put in the analysis folder? I want to try to move things out of nbtools and this seems like a good can

tests/test_timeseries.py
```diff
@@ -8,16 +8,39 @@
...
 class TestTimeSeries(unittest.TestCase):
     """Unit test class."""
+    def setUp(self):
+        input_file = os.path.join(_TEST_DATA, "TimeSeriesDemo.csv")
+        self.input_df = pd.read_csv(input_file, index_col=["TimeGenerated"], usecols=["TimeGenerated","TotalBytesSent"])
+
+    def test_timeseries_anomalies_stl(self):
+        out_df = timeseries_anomalies_stl(data=self.input_df)
+
+        self.assertIn("residual", out_df.columns)
```
ianhelle, line 37: Given that you have a known data set can you check for actual expected values produced - e.g. the number of anomalies or something like that? I know the code is really simple but if something changes in statsmodel and everything is zeros, we'd never know.

msticpy/nbtools/timeseries.py
ianhelle: I can also add my time region extraction functions to this.

setup.py
ashwin-patil: Zscore is available in older version of scipy but STL method was introduced statsmodels - 0.11.0 so will need upgrade. I will make changes for scipy and also add it to conda-reqs. 
- https://www.statsmodels.org/dev/release/version0.11.html
- https://github.com/statsmodels/statsmodels/pull/5926

docs/source/visualization/TimeSeriesAnomalies.rst
ashwin-patil: fixed errors.

tests/test_timeseries.py
ashwin-patil: Thanks . Added check for anomalies

requirements.txt
ashwin-patil: Added check. test passing now.

msticpy/nbtools/timeseries.py
ashwin-patil: thanks. added accordingly.

msticpy/nbtools/timeseries.py
ashwin-patil: moved function under analysis root.

PR #71, opened 2020-06-12
Msticpy output fix

ianhelle:
Mostly fixes to remove dependency on Output widget.
Created new versions of AlertSelector and SelectString - SelectAlert, SelectItem - and deprecated old ones.
Also added IPython displayable capabilities to entities, security alerts and the widgets themselves.
Moved a couple of package versions down to match AzureNotebooks defaults to improve install speed.


msticpy/nbtools/nbdisplay.py
```diff
@@ -203,59 +230,123 @@ def display_logon_data(
          explicitly specify os_family (Linux or Windows)
          (the default is None)
 
+    """
+    display(format_logon(logon_event, alert, os_family))
+
+
+@export
+def format_logon(
+    logon_event: Union[pd.DataFrame, pd.Series],
```
petebryan, line 239: This looks like it only processes Linux Audit logons events and not Syslog events - you might want to reference this in the doc string.

ianhelle: Added note to say only Windows logons supported currently (added in later branch)

PR #74, opened 2020-06-25
Allow specifying yaml files & TZ issue fix

petebryan:
Fixes:
 - Updated wsconfig so that when you call WorkspaceConfig(config_file=path) to a msticpyconfig.yaml it doesn't throw a JSON decode error and instead loads that file as a custom config.
- Updated cluster_syslog_logons_df to handle mix of TZ and TZ agnostic data.
- Reduced size of test data set to be more efficent.


petebryan:
No longer required after branch merges.

PR #76, opened 2020-06-25
Merge current working branches

petebryan:


msticpy/common/wsconfig.py
```diff
@@ -92,7 +93,13 @@ def __init__(self, config_file: Optional[str] = None, workspace: str = None):
                     "\n".join(_NO_CONFIG_WARN).format(config_file=config_file)
                 )
         self._config_file = config_file
-        self._config.update(self._read_config_values(config_file))
+        config = self._read_config_values(config_file)
+        if config is not None:
```
ianhelle, line 97: Not sure about this. config_file is meant to be a config.json. I guess the doc string is misleading - I originally meant for this to be a way to override msticpyconfig and force load of a json file. I guess if it works specifying msticpyconfig.yaml as config_file maybe we should keep it.

msticpy/common/wsconfig.py
```diff
@@ -92,7 +93,13 @@ def __init__(self, config_file: Optional[str] = None, workspace: str = None):
...
+            self._config.update(config)
+        else:
+            os.environ["MSTICPYCONFIG"] = config_file
```
ianhelle, line 100: I wonder if this might cause some issues. E.g. if I already have a MSTICPYCONFIG env var but wanted to force use of a local msticpyconfig just for workspace settings. It's probably OK to leave it as you have it. It will be a pretty rare use case anyway and it's probably a nightmare to support simultaneous use of multiple msticpyconfig.yamls

PR #82, opened 2020-07-08
Ianhelle/2020 07 07 timeline datetime fix

ianhelle:
Fix for new datetime dtypes to correctly display datetime in timelines


ianhelle:
Extracting this change to another branch

PR #85, opened 2020-07-16
Error in pkg_config validate when no config sections are populated.

ianhelle:
Exception thrown when a checked section is missing

### DIFF
--- a/msticpy/common/pkg_config.py
+++ b/msticpy/common/pkg_config.py
@@ -336,12 +336,15 @@ def _print_validation_report(mp_errors, mp_warn):
 
 def _validate_azure_sentinel(mp_config):
     mp_errors = []
+    mp_warnings = []
     as_settings = mp_config.get(_AZ_SENTINEL, {})
     if not as_settings:
         mp_errors.append("Missing or empty 'AzureSentinel' section")
+        return mp_errors, mp_warnings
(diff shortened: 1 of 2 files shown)


msticpy/common/pkg_config.py
```diff
@@ -347,10 +347,10 @@ def _validate_azure_sentinel(mp_config):
         if ws == "Default":
```
petebryan, line 347: These changes still don't catch the fact an exception is caused by calling .items() on None if Azure Sentinel section doesn't have any elements properly populated. There is a check at line 343 that checks that ws_settings exists and adds an error stating workspaces is missing but it then proceeds to try and process the section anyway. We can solve this easily by wrapping the second half in a else statement:
```
 if not ws_settings:
        mp_errors.append("Missing or empty 'Workspaces' key in 'AzureSentinel' section")
        no_default = True
    else:
        for ws, ws_settings in ws_settings.items():
            if ws == "Default":
                no_default = False
            ws_id = ws_settings.get("WorkspaceId")

```

ianhelle: Yep, good catch. I think the error I saw was in the check_provider_settings function but both have the same bug.
(as does the line above this one).

juliango2100:
Tested and it works as expected.  Approved and merged

PR #91, opened 2020-08-24
VirusTotal API v3

aramirezmartin:
Added new class VTLookupV3, which makes use of VirusTotal API v3 and VTGraph . With the new API, it is possible to obtain relationships for different entity types. Then, users can render those relationships using VTGraph.

Added a documentation notebook to explain new functionalities implemented.


ianhelle:
@aramirezmartin - hey thanks for the PR. Looks really cool!. I'm about to head out for vacation for a 1.5 weeks so may not get chance to review properly before I get back.

ianhelle:
Content of PR moved to #97

PR #95, opened 2020-09-05
Remove typing - Redundant Requirement

roopeshvs:
> NOTE: in Python 3.5 and later, the typing module lives in the stdlib, and installing this package has NO EFFECT, because stdlib takes higher precedence than the installation directory.

As msticpy requires Python 3.6 and above, this is now a redundant requirement.

Refer: https://pypi.org/project/typing/


ianhelle (approved):

ianhelle:
@roopeshvs - thanks for the fix! Much appreciated.

PR #105, opened 2020-09-29
Suggestions for refactoring code.

ianhelle:
I've also done a bit of tidy-up in the Keyvault code moving KeyVaultSettings into its own module.


petebryan:
Made changes in branch rather than attempting merge.

PR #109, opened 2020-10-08
Az Auth, and Sentinel

petebryan:
Fixes:
- Fixed issue of renamed Python SDK clients from Azure

New Features:
- New common Azure auth methods (common.azure_auth, common.azure_auth_core)
- Updated Azure Data and KeyVault fucntions to work with this
- Added new functions to handle calling Sentinel APIs
- Updated unit tests and docs.


msticpy/data/azure_sentinel.py
```diff
@@ -0,0 +1,209 @@
...
+            for wrkspace in workspaces:
+                name = wrkspace.split("/")[-1]
+                workspaces_dict.update({name: wrkspace})
+
+            return workspaces_dict
+
+        print(f"No Azure Sentinel workspaces in {sub_id}")
+        return {}
+
+    def get_hunting_queries(self, res_id) -> pd.DataFrame:
```
ianhelle, line 81: The resource ID is especially klunky and ripe for user errors.
It might be good to have something that sets the current workspace for the class. This would store sub_id, resource_group, workspace name. Then you could just build the resource string without having to supply a parameter.
You could still have the param available as an override for the default.

Perhaps in the connect function we have optional params for subID and workspace name or ID. Then we set those to self. properties. All of the other methods could then reference these values.

ianhelle, line 81: Surprised that you didn't get any warnings about these being instance methods but not referencing self.

msticpy/sectools/vtlookupv3.py
```diff
@@ -15,7 +15,7 @@ class MsticpyVTNoDataError(Exception):
 
 
 class MsticpyVTGraphSaveGraphError(Exception):
-    """Could not save VT Garph."""
```
ianhelle, line 18: I did some updates to this to get rid of some linting/black warnings - you might want to check that I fixed this typo when merging

docs/notebooks/DataObfuscation.ipynb
```diff
@@ -1716,4 +1716,4 @@
  },
  "nbformat": 4,
  "nbformat_minor": 4
-}
```
ianhelle, line 1719: Seems odd that these changes (there's another later on) made their way in. Not 100% sure whether newline at the end of the file is needed.

tests/test_azure_sentinel.py
```diff
@@ -0,0 +1,22 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+"""Azure Sentinel unit tests."""
+from pytest import raises
+
+from msticpy.data.azure_sentinel import AzureSentinel
```
ianhelle, line 9: We should try to create mocks for this and AzureData - feels like we should be able to get pretty good coverage (albeit with mocks and not the real API)

tools/config2kv.py
```diff
@@ -60,9 +60,9 @@
 
 
 def _read_config_settings(conf_file):
-    sys_config = conf_file = os.environ["MSTICPYCONFIG"]
-    if not conf_file:
-        conf_file = sys_config
+    # sys_config = os.environ["MSTICPYCONFIG"]
```
ianhelle, line 63: why are these commented out?

ianhelle (changes requested):
There are a couple of typos and some questions. Happy to chat about what to do about some of them. Most of the comments that relate to big things (like mocks for azure_data, AzureSentinel tests) are more "we should do this in the future". Not sure about the AzureSentinel API. Maybe we should release without announcing this part and fiddle about with it for the next release.
Nice work though!

msticpy/data/azure_sentinel.py
```diff
@@ -0,0 +1,209 @@
...
+        auth_methods: List = None,
+        silent: bool = False,
+    ):
+        """Authenticate with the SDK & API."""
+        super().connect(auth_methods=auth_methods, silent=silent)
+        self.token = _get_token(self.credentials)
+        self.res_group_url = None
+        self.prov_path = None
+
+    def get_sentinel_workspaces(self, sub_id: str) -> Dict:
```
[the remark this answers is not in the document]
petebryan: So its not auto_populated but as a subclass of AzureData you can call:
asz = AzureSentinel()
asz.connect()
sub = asz.get_subscriptions()
asz.get_sentinel_workspace(sub['Subscription ID'].iloc[0])

I can look at defaulting to a subscription includes in config.json if its present.

msticpy/data/azure_sentinel.py
petebryan: Not sure I agree with this, users are going to be accessing these elements in either one of two ways - the first is programmatically where passing a resource ID is easy especially as get_sentinel_workspaces returns it for you:
ws = azs.get_sentinel_workspaces(sub_id)
azs.get_hunting_queries(ws['MyWorkspace'])

The other way is people copy and pasting values into the Notebook, probably from the Azure portal in which case it would see one copy and paste of the resource ID is safer and less error prone than 3 separate copy and pastes.

I'll add support for passing the individual elements through if you like, an option and an option to default to try and get them from config.

Ill think on the idea of passing stuff at connect, my gut feeling is that this isn't great as it implies these value are then set for everything in the class which we don't want as most Sentinel customers have multiple workspaces in multiple subscriptions so will likely want to use this to query multiple of them. If I add an option that if you don't provide arguments for the functions it attempts to get them from config files it will effectively achieve the same thing as if you do have a single workspace you can just let it pull from config.

petebryan: They reference self.token (and with the updates I've added self._check_config as well now)

tools/config2kv.py
petebryan: This script wasn't working - I had to comment out some lines to get it to work and forgot to make proper changes. Ill do that now.

tests/test_azure_sentinel.py
petebryan: I've done Azure Sentinel tests, I'll come back and increase Azure Data coverage at a later date.

ianhelle (approved):

PR #117, opened 2020-11-12
Improved vtlookup3 performance

aramirezmartin:


ianhelle:
@aramirezmartin - Andres, first this is super cool. Thanks for submitting.
Question: are you able to see the details of the build run? I'm guessing not since the pipeline is attached to an internal repo. I need to work out some way of fixing that since GitHub only shows an unhelpful "cmd.exe exited with 1 error code".

A couple of requests:
1. Can you run black and (ideally all the linters but at least prospect, which mostly runs everything else). Our very small dev doc page is here https://github.com/microsoft/msticpy/wiki/Contributor-guidelines
2. In the notebook I made a change to convert the display of serial timestamps into pandas datetimes - I think it would be better to do this in the module though. No hurry for this. Can be a future PR.

I checked out the branch and did the linting fixes (also one of them spotted a function with an unused parameter `limit` in lookup_ioc_relationships. I'm assuming that you meant to pass this to the async version, so I've made that change.

I think I'm borrow your sync-async bridge pattern for other things 👍. Not yet super comfortable with async in Python but this looks very cool.

ianhelle (approved):

ianhelle:
@aramirezmartin - Andres, first this is super cool. Thanks for submitting.
Question: are you able to see the details of the build run? I'm guessing not since the pipeline is attached to an internal repo. I need to work out some way of fixing that since GitHub only shows an unhelpful "cmd.exe exited with 1 error code".

A couple of requests:
1. Can you run black and (ideally all the linters but at least prospect, which mostly runs everything else). 
2. In the notebook I made a change to convert the display of serial timestamps into pandas datetimes - I think it would be better to do this

PR #120, opened 2020-11-18
Updating VTLookup3 notebook and module

ianhelle:
Updated VTLookupV3 notebook to add a couple of capabilities.
Added get_object function and some date formatting to vtlookupv3.py.
Added vtlookupv3 module to API docs.


petebryan (approved):
Looks good, I've just added a commit to fix a broken link that was caused by an update to an external site.

PR #122, opened 2020-11-25
Pebryan/2020 10 14 headless work

petebryan:
New features:

- Azure Blob Storage functionality
- New Azure Sentinel API functionality (get incidents, update incidents, add comments)
- New Az Cli and MSI auth support for KQL data provider
- Test and docs to support these features


msticpy/data/azure_blob_storage.py
```diff
@@ -0,0 +1,260 @@
...
+            )  # type:ignore
+        except ResourceExistsError as err:
+            raise CloudError(f"Container {container_name} already exists.") from err
+        properties = new_container.get_container_properties()
+        container_df = _parse_returned_items(
+            [properties], ["encryption_scope", "lease"]
+        )
+        return container_df
+
+    def blobs(self, container_name: str) -> pd.DataFrame:
```
ianhelle, line 93: Since you could return None, this will prob generate a mypy error - return type should be Optional[pd.DataFrame

msticpy/data/azure_blob_storage.py
```diff
@@ -0,0 +1,260 @@
...
+
+        Returns
+        -------
+        pd.DataFrame
+            Details of the blobs.
+
+        """
+        container_client = self.abs_client.get_container_client(container_name)  # type: ignore
+        blobs = list(container_client.list_blobs())
+        if blobs:
```
ianhelle, line 110: can simplify to 
return _parse_returned_items(blobs) if blobs else None

msticpy/data/azure_data.py
```diff
@@ -226,22 +226,15 @@ def get_resources(  # noqa: MC0001
...
-        # pylint: disable=unnecessary-comprehension
-        try:
-            [r for r in self.resource_client.resources.list()]  # type: ignore
-        except AttributeError:
-            self._legacy_auth("resource_client", sub_id)
-
+        resources = []  # type: List
         if rgroup is None:
-            # pylint: disable=line-too-long
-            resources = [resource for resource in self.resource_client.resources.list()]  # type: ignore
```
ianhelle, line 237: What was wrong with the list comprehension? - they are much faster than for loops. Sourcery prefers them

msticpy/data/azure_data.py
```diff
@@ -226,22 +226,15 @@ def get_resources(  # noqa: MC0001
...
+            for resource in self.resource_client.resources.list():  # type: ignore
+                resources.append(resource)
         else:
-            resources = [
```
ianhelle, line 239: ditto

requirements.txt
```diff
@@ -11,6 +11,7 @@ azure-mgmt-monitor>=1.0.1
 azure-mgmt-network>=2.7.0
 azure-mgmt-resource>=2.2.0
 azure-mgmt-subscription>=0.2.0
+azure-storage-blob>=12.5.0
```
ianhelle, line 14: I'm going to kick a few of these out of requirements.txt. This is ok to be optional with the other azure components right?

tests/test_azure_sentinel.py
```diff
@@ -10,6 +10,7 @@
 import pandas as pd
 import pytest
 import responses
+from azure.core.exceptions import ClientAuthenticationError
```
ianhelle, line 13: This should probably go in data folder as well right?

tests/test_abs.py
```diff
@@ -0,0 +1,185 @@
+# -------------------------------------------------------------------------
```
ianhelle, line 1: Could you put this in the data subfolder and give it a long descriptive name? Not as if we need to type it. The pattern I use is test_name_of_module_being_tested.py - makes it easier to work out what is doing what.

tools/config2kv.py
```diff
@@ -76,7 +76,7 @@ def _read_config_settings(conf_file):
     os.environ["MSTICPYCONFIG"] = conf_file
     config.refresh_config()
     kvlt_settings = KeyVaultSettings()
-    os.environ["MSTICPYCONFIG"] = sys_config
+    # os.environ["MSTICPYCONFIG"] = sys_config
```
ianhelle, line 79: Didn't we change this before? Looks redundant from line 76

ianhelle (dismissed):
Nice job! I have a few comments scattered around

requirements.txt
petebryan: yep thats good.

msticpy/data/azure_data.py
petebryan: So that's what I thought but pylint throws this whenever I used list comprehension here:
image

ianhelle (approved):

msticpy/data/azure_data.py
ianhelle: In this particular line it is unnecessary because it looks like self.resource_client.resources.list() is already a list (so the list comp is not doing anything)

ianhelle (approved):

PR #124, opened 2020-11-30
Adding credscan suppressions and changing a few dummy passwords to pr…

ianhelle:
Fixing current credscan warnings prior to running BFG to fixup past commits


ianhelle:
This PR mistakenly had all of the re-written commits for credscan issue. Deferring this for now

PR #126, opened 2020-12-02
Adding credscan suppressions and changing a few dummy passwords to pr…

ianhelle:
Phase 1 of credscan FP fix


ianhelle:
Need this to be merged to prevent further security false alarms

PR #130, opened 2021-01-13
Ianhelle/user defaults 2021 01 11

ianhelle:
Allow user to specify startup defaults (which providers are loaded/initialized) in msticpyconfig.yaml


ianhelle:
This is now merged into #136

PR #167, opened 2021-06-04
Resource Graph Explorer (local branch)

petebryan:
This is a local branch of PR 163 - created to work with our Pipeline.


msticpy/sectools/domain_utils.py
```diff
@@ -226,7 +226,7 @@ def in_abuse_list(self, url_domain: str) -> Tuple:
             )
         except Exception:  # pylint: disable=broad-except
             result = False
-            x509 = None
+            x509 = None  # type: ignore
```
ianhelle, line 229: I think I fixed this in the data_viewer branch as well, maybe a bit more elaborately (ISTR importing the Certificate type)

tests/data/drivers/test_drivers.py
```diff
@@ -11,7 +11,7 @@
 
 from msticpy.data.drivers.mdatp_driver import MDATPDriver
 from msticpy.data.drivers.security_graph_driver import SecurityGraphDriver
-
+from msticpy.data.drivers.resource_graph_driver import ResourceGraphDriver
```
ianhelle, line 14: You might want to surround this with a try/except ImportError.
Not strictly necessary but I did this in other places - and then if the import fails have a pytest.mark.skip that causes the whole test to be skipped.
The advantage to this is that all the tests pass or are skipped even if all the extras are not installed. Otherwise you get a test error.
Looks like I didn't do this consistently - maybe I ended up thinking that test errors were OK and only did it where there were test files with some cases that worked with core install and some didn't.
If you want to see an example look at data/test_query_source and look for the import statements and @pytest.mark.skipif(not _SPLUNK_IMP_OK, reason="Partial msticpy install")
lines.
I don't think it's worth doing here - I didn't do it for the azure_data test and a bunch of others.

ianhelle (approved):

msticpy/sectools/domain_utils.py
petebryan: Seen that, Ill merge with your data_viewer fix

ianhelle (approved):

PR #171, opened 2021-06-08
Pebryan/2021 6 7 grey noise updates

petebryan:
Updated GreyNoise elements with correct naming convention.
Added custom MSTICPy user agent to  TI providers for usage tracking.


msticpy/common/utility.py
```diff
@@ -662,3 +663,8 @@ def _merge_dicts(dict1: Dict[Any, Any], dict2: Dict[Any, Any]):
             d_val = dict1[key]
         out_dict[key] = d_val
     return out_dict
+
+
+def get_user_agent():
```
ianhelle, line 668: Maybe this would be better as a constant (although we might, in future want to build it dynamically). I think it should also be named more specifically - like MSTICPY_USER_AGENT. I think the generic name might be confusing.
Then we'd just from utility import MSTICPY_USER_AGENT wherever we needed to use it.
I think we should (in time) add it to all our requests-based modules.
(aside - I think we should also replace requests with httpx - but that's another story)

msticpy/sectools/tiproviders/http_base.py
```diff
@@ -215,6 +215,8 @@ def _substitute_parms(
                 key: val.format(**req_params) for key, val in src.headers.items()
             }
             req_dict["headers"] = headers
+        if "User-Agent" not in req_dict["headers"].keys():
+            req_dict["headers"]["User-Agent"] = get_user_agent()
```
ianhelle, line 219: Ah cool - it is automatically going into many of our providers!

ianhelle (approved):

msticpy/common/utility.py
petebryan: Good point, I've converted to a constant

ianhelle: Oh - maybe my main point was lost. I was saying that we should call it MSTICPY_USER_AGENT - or something along those lines.
Maybe:
```python
def _get_mp_ua():
   return ....your code

MSTICPY_USER_AGENT = _get_mp_ua()
```
I think get_user_agent is too generic - more like something you'd find as a pivot function.

PR #176, opened 2021-06-23
Timeline duration control

ianhelle:
Added new timeline control - displaying start-end durations of grouped events.
Tidying up some of the code in timeline.py
Moving the pd accessors to timeline_pd_accessor.py
Updating notebook EventTimeline.ipynb and documentation EventTimeline.rst


msticpy/nbtools/timeline_duration.py
```diff
@@ -0,0 +1,248 @@
...
+    range_tool: bool = True
+    xgrid: bool = True
+    ygrid: bool = False
+    hide: bool = False
+    color: str = "navy"
+    ylabel_cols: Iterable[str] = attr.Factory(list)
+    ref_events: Optional[pd.DataFrame] = None
+    ref_col: Optional[str] = None
+    ref_times: Optional[List[Tuple[datetime, str]]] = None
+
```
2xyo, line 60: `range_tool` and `source_columns` are missing :)

ianhelle: Thx. source_columns is missing by design. Since the way I create the durations is by grouping on the group_by column(s) and then finding the max and min of the timestamps, all of the other field values for individual events disappear. 
They kind of have to be ignored really since I'm trying to assign Tooltip values to the entire duration - so only values that are common to the grouped duration can be shown.

range_tool is an omission. I think I've fixed in an update with tests that I'm about to push.

petebryan (approved):
This is great. Love the restructure makes it much cleaner.

PR #178, opened 2021-06-25
Process tree updates to handle more generic data

ianhelle:
Note: the cf3b346 commit (Jun 25) has the wrong text in it (not sure how)
This PR updates process_tree_utils to handle more generic data. You should now be able to specify
the schema for input data using either the ProcSchema class or a dict with the same keys as ProcSchema.

Also improved unit test coverage by adding debug=True to build tree, added test for specifying schema with dict
and a locally-running test for the Bokeh chart building.


marcohenriques:
Hi @ianhelle , as you're working on this, the [Progress widget](https://github.com/microsoft/msticpy/blob/ianhelle/process-tree-update-2021-06-25/msticpy/nbtools/nbwidgets.py#L1453) I think has an outdated attribute for the ipywidgets version the project is using. The setter for the attribute `visible` for the widgets is no longer available, which makes this progress bar to be always displayed. I believe the new way is by using the layout, something like:
```
check.layout.visibility = 'hidden'
or
check.layout.display = 'none'

```
[reference](https://github.com/jupyter-widgets/ipywidgets/issues/674#issuecomment-233566026)

petebryan (approved):

PR #179, opened 2021-06-29
Ianhelle/mp pivot suppress exceptions 2021 06 28

ianhelle:
Added mechanism to stop standard display of msticpy user
exceptions. Context manager in MsticpyUserError:
``` python
from msticpy.common.exceptions import MsticpyUserError
with MsticpyUserError.no_display_exceptions():
   # do stuff that may cause exceptions to be displayed.
```
Note this doesn't suppress the exception, just stops exception being output to a notebook.
This is needed because MsticpyUserError and subclasses will display to IPython or stdout when
created, even if the exception is subsequently caught.

### DIFF
--- a/msticpy/sectools/geoip.py
+++ b/msticpy/sectools/geoip.py
@@ -219,10 +219,7 @@ Alternatively, you can pass this to the IPStackLookup class when creating it:
         super().__init__()
 
         self.settings = _get_geoip_provider_settings("IPStack")
-        if api_key:
-            self._api_key = api_key
-        else:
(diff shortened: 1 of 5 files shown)


msticpy/sectools/geoip.py
```diff
@@ -219,10 +219,7 @@ def __init__(self, api_key: Optional[str] = None, bulk_lookup: bool = False):
         super().__init__()
```
ianhelle, line 219: Note: this was just a sorcery refactoring - no logic change here

petebryan (approved):

PR #191, opened 2021-08-05
fix img src paths for broken preview images in readme

ashwin-patil:
- fix img src paths to local to preview images correctly.


ianhelle:
I changed the URLs to the absolute path a while ago so that they'd appear correctly in PyPI but it doesn't look as though this displays them anyway.

PR #194, opened 2021-08-20
Ianhelle/update pipeline yam 2021 06 09

ianhelle:


ianhelle:
Moving updates to azure-pipelines.yaml to clean branch

PR #201, opened 2021-09-10
Ianhelle/azure pipelines yaml 2021 09 09

ianhelle:
Adding azure pipelines and updating github actions yamls


azure-pipelines.yml
```diff
@@ -0,0 +1,297 @@
...
+    strategy:
+      matrix:
+        linux:
+          imageName: "ubuntu-latest"
+          python.version: '3.6'
+        mac:
+          imageName: "macos-latest"
+          python.version: '3.6'
+        windows-py36:
+          imageName: "windows-latest"
```
petebryan, line 45: Do we want to add windows 3.8?

petebryan (approved):

azure-pipelines.yml
ianhelle: We could. I think I remember a funny issue with bokeh on Win + 3.8 that caused a test to fail because of missing attribute.

PR #209, opened 2021-09-18
Azure CLI credentials check

ianhelle:
Added check_cli_credentials to azure_auth_core.py
Added check to nbinit.py to check and print status of Azure CLI login (if AzureCLI is installed)
Changed azure_ml_tools.py to print out the installed rather than required version of MSTICPy
Added unit test for check_cli_credentials in test_azure_auth_core.py
Updated version to 1.4.4


petebryan (approved):
Looks great, super cool feature!

PR #211, opened 2021-10-06
Pebryan/2021 8 18 incident triage

petebryan:
Features added:

- Entity Graph feature to create, build, and visualize an entity graph:
image
This includes updates to entities and other supporting objects to enable this feature.
Also added Pandas accessors for this feature.
- Support for new Azure Sentinel Preview API for entities and alert elements.


msticpy/datamodel/entities/account.py
```diff
@@ -187,4 +205,7 @@ def _create_from_event(self, src_event, role):
         # DisplayName (type System.String)
         "DisplayName": None,
         "ObjectGuid": None,
+        "TimeGenerated": None,
```
ianhelle, line 208: Are these part of the Account schema?
Seems odd to have timestamps associated with a persistent object like an account

msticpy/datamodel/entities/alert.py
```diff
@@ -71,25 +83,81 @@ def __init__(self, src_entity: Mapping[str, Any] = None, **kwargs):
         self.DisplayName: Optional[str] = None
         self.CompromisedEntity: Optional[str] = None
         self.Count: Any = None
-        self.StartTimeUtc: datetime = datetime.min
```
ianhelle, line 74: Are we breaking V3 compatability here.
Seems that Start and EndttimeUtc are the official properties. https://dev.azure.com/msazure/One/_git/Rome-Detection-AlertContracts?path=/doc/AlertV3.md&_a=preview

Perhaps we should have both - i.e. allow the same value to be set from either name and return either the Utc or non-Utc suffixed versions.

msticpy/datamodel/entities/alert.py
```diff
@@ -4,12 +4,15 @@
 # license information.
```
ianhelle, line 4: There's a bit of an issue here (that I've helped to create). We should perhaps tidy this up and sweep up the SecurityAlert class at the same time.
In the V3 schema there is the SecurityAlert class and there is also an Alerts class - the latter was meant to be used to hold collections of alerts when you have an incident or an alert with related alerts (although seemed a bit of dumb kind of class to have IMHO - better to have just had edges).

We should probably consolidate into a single Alert V3 class that follows the V3 schema plus any Azure Sentinel extras.
And keep an Alerts glue entity. 
https://dev.azure.com/msazure/One/_git/Rome-Detection-AlertContracts?path=/src/AlertContracts/V3/Entities/Alerts.cs
vs.
https://dev.azure.com/msazure/One/_git/Rome-Detection-AlertContracts?path=/doc/AlertV3.md&_a=preview

Actually the former doesn't even appear in the docs so maybe we can just drop Alerts and consolidate the Alert one.
In the V3 schema SecurityAlert is not an entity but a separate thing. Not sure we should end up making it an entity.
Maybe we should put the Alert and Incident classes in a peer folder like soc or something.

msticpy/datamodel/entities/file.py
```diff
@@ -141,19 +148,27 @@ def _create_from_event(self, src_event, role):
         # Sha256Ac (type System.String)
         "Sha256Ac": None,
         "FileHashes": (list, "FileHash"),
+        "TimeGenerated": None,
```
ianhelle, line 151: If these are common to all entities we might want to think about implementing this a bit more efficiently.
Have a base ENTITY_SCHEMA dict in Entity and define individual _entity_schema as
{
   **(Entity.ENTITY_SCHEMA),
   "prop1": type
   ...
}
WDYT?

msticpy/datamodel/entities/process.py
```diff
@@ -122,29 +115,19 @@ def _create_from_event(self, src_event, role):
...
-            self.euid = src_event["euid"] if "euid" in src_event else None
-            self.effective_group = (
-                src_event["effective_group"] if "effective_group" in src_event else None
-            )
-            self.egid = (
-                src_event["effective_group"] if "effective_group" in src_event else None
-            )
-            self.cwd = src_event["cwd"] if "cwd" in src_event else None
-            self.name = src_event["cwd"] if "cwd" in src_event else None
+            self.success = src_event.get("success")
```
ianhelle, line 118: Brilliant! I never realized the pd.Series support dict-like get()
Wow!

msticpy/datamodel/entities/security_group.py
```diff
@@ -63,11 +63,19 @@ def description_str(self):
         """Return Entity Description."""
         return self.DistinguishedName or self.__class__.__name__
 
+    @property
```
ianhelle, line 66: Also if we implemented this in the Entity class, we would only need to implement in entity sub-classes where it was returning something different to self.__class__.__name__

msticpy/datamodel/entities/soc/incident.py
```diff
@@ -0,0 +1,169 @@
...
+from ...entities.entity import Entity, camelcase_property_names
+from ...entities.alert import Alert
+
+__version__ = VERSION
+__author__ = "Pete Bryan"
+
+
+# pylint: disable=invalid-name, too-many-instance-attributes
+@export
+class Incident(Entity):
```
ianhelle, line 22: Should this be an entity?
I guess technically not but maybe it helps with the edges and stuff.
One thing that we should do for this and the Alert class is - whenever we add an entity or alert to part of the parents collection (e.g. in this case alerts) we should add an edge to the alert.
That would also make the graph generation easier.
Same thing with alerts and it entities member.

msticpy/datamodel/entities/soc/incident.py
```diff
@@ -0,0 +1,169 @@
+# -------------------------------------------------------------------------
```
ianhelle, line 1: should this be a subfolder of entities or have a separate soc folder under datamodel?
We could put stuff like Alert, Incident, and whatever else (bookmark, observations,...) in there

msticpy/vis/entity_graph_tools.py
```diff
@@ -0,0 +1,559 @@
...
+            Type="analystnote",
+            TimeGenerated=datetime.now(),
+        )
+        if attached_to:
+            if isinstance(attached_to, str):
+                attached_to = [attached_to]
+            for link in attached_to:
+                self.add_link(name, link)
+
+    def add_link(self, source: str, target: str):
```
ianhelle, line 240: Possibly not show-stopping but the usual terminology for this is add_edge, delete (or remove)_edge.
I think we have this (at least add_edge in the entities.entity_graph module).

msticpy/vis/entity_graph_tools.py
```diff
@@ -0,0 +1,559 @@
...
+        if alrt["Entities"]:
+
+            alrt_ents += [ent.__hash__() for ent in alrt["Entities"]]
+    for ent in ents:
+        if ent.__hash__() in alrt_ents:
+            ents.remove(ent)
+    return ents
+
+
+@pd.api.extensions.register_dataframe_accessor("mp_incident_graph")
```
ianhelle, line 413: I think I mentioned this before somewhere. 
I was hoping to start consolidating our pandas extensions so that we don't have one for every features. The plotting ones are easiest - since they can be nicely grouped under an mp_plot. I think I did this for the matrix plot - it doesn't have its own accessor.
We should deprecate the other (timeline, etc) and remove them at some point.
The only reason to keep a separate accessor like this is if you need something very specific to the entity_graph, ,ike the build_graph function.
Are you OK with removing this and just relying on mp_plot?

requirements.txt
```diff
@@ -9,7 +9,7 @@ azure-identity>=1.5.0
 # azure-mgmt-keyvault>=2.0.0  # azure
 # azure-mgmt-monitor>=1.0.1  # azure
 # azure-mgmt-network>=2.7.0  # azure
-# azure-mgmt-resource>=2.2.0  # azure
+# azure-mgmt-resource>=16.1.0  # azure
```
ianhelle, line 12: That's a big version jump - they must have been very busy!

ianhelle:
That's just some awesome work @petebryan. Just bowled over by the attention to detail in so many areas! Brilliant (my comments, notwithstanding)

msticpy/datamodel/entities/account.py
petebryan: https://github.com/microsoft/msticpy/issues/215

ianhelle (approved):

PR #231, opened 2021-11-15
Ianhelle/kusto driver fixes 2021 11 12

ianhelle:
Adding support for MSAL/Azure integrated auth.
Added unit tests for this (majority of the new code)
Fixing a couple of bugs uncovered by Kusto config settings/test in pkg_config and nbinit


tests/testdata/kusto/kusto_queries_mstic_ia.yaml
```diff
@@ -101,7 +101,7 @@ sources:
     description: Lists all processes with a command line containing a string
     metadata:
       data_families: [scrubbeddata]
-      cluster: https://wcdscrubbedservice.kusto.windows.net
+      cluster: https://mstic.kusto.windows.net
```
petebryan, line 104: Shall we use msticpy for the cluster name here, just in case one day we have a mstic cluster for real. Not a big issue just a thought.

msticpy/data/drivers/kusto_driver.py
```diff
@@ -19,12 +19,9 @@
...
 
-_KCS_TEMPLATE = (
-    "azure_data-Explorer://"
-    "tenant='{tenant}';clientid='{clientid}';clientsecret='{clientsecret}';"
-    "cluster='{cluster}';"
-    "database='{database}'"
-)
+_KCS_CODE = "code;"
+_KCS_APP = "tenant='{tenant}';clientid='{clientid}';clientsecret='{clientsecret}';"
+_KCS_TEMPLATE = "azure_data-Explorer://{auth}cluster='{cluster}';database='{database}'"
```
petebryan, line 24: Its probably not worth taking a dependency but this could be useful - https://github.com/Azure/azure-kusto-python/blob/35e23cef5d246328245a6e34c642db818aec1b19/azure-kusto-data/azure/kusto/data/client.py

petebryan (approved):

tests/testdata/kusto/kusto_queries_mstic_ia.yaml
ianhelle: Not sure it's a big deal - I don't think there and this doesn't give anyone access. Just thought I'd make it sound a bit more generic

msticpy/data/drivers/kusto_driver.py
ianhelle: This is interesting. I'm thinking of looking at it to potentially replace Kqlmagic.
It had problems initially (like no LogAnalytics) but might have worked through them now

Issue #233, opened 2021-11-19
Regularize NBWidgets


ianhelle:
Current the widgets have a variety of parameters and properties that don't quite match and don't follow the ipywidgets property names.
This mainly applies to the selection widgets.

To Dos:
- All widgets should expose a layout property - which is the overall HBox/VBox container widget
- All widgets should expose a value property - ideally a get and set
- ipywidgets in the control should be exposed as public attributes so that they can be manipulated if necessary
- Use the following parameter conventions:
   - description for title
   - options for select options
   - width (if appropriate)
   - height (if appropriate)
- for selection widgets give a way to programmatically select the selected option.

ianhelle:
We should fix these two issues in RegisteredWidget (in nbwidgets):
Change globals to frame hack. Fix naming of "ns"
```python
    def __init__(
        self,
        id_vals: Optional[List[Any]] = None,
        val_attrs: Optional[List[str]] = None,
        nb_params: Optional[Dict[str, str]] = None,
        ns: Dict[str, Any] = globals(),
        ns: Dict[str, Any] = globals(),  # pylint: disable=invalid-name
```

PR #239, opened 2021-12-17
Ianhelle/mde proctree fixes 2021 12 16

ianhelle:
Fixes for MDE process tree to deal with date strings. Added same fix to timelines and win/linux process tree builders.
Change default behavior of kql_driver to use Azure integrated authn
Changed a bunch of UI/messages to "Microsoft Sentinel"
Updated versions of black and pylint in pre-commit yaml


msticpy/nbtools/nbwidgets.py
```diff
@@ -80,7 +80,7 @@ def __init__(
         id_vals: Optional[List[Any]] = None,
         val_attrs: Optional[List[str]] = None,
         nb_params: Optional[Dict[str, str]] = None,
-        ns: Dict[str, Any] = globals(),
+        ns: Dict[str, Any] = globals(),  # pylint: disable=invalid-name
```
petebryan, line 83: couldn't we quite easily re-name this variable. Something like 'name_space'?

petebryan (approved):

msticpy/nbtools/nbwidgets.py
ianhelle: I've added this as comment in this issue #233

PR #262, opened 2022-01-06
20220105 hackathon blob storage conn string

lucky-luk3:
Added the ability to connect to a blob storage via connection string.


petebryan:
Thanks @lucky-luk3 this looks great! There are a couple of failing checks for black and pylint - if you can take a quick look and update. 

I think it should just be the case of running `black azure_blob_storage.py`

Let me know if you have any issues with that and

lucky-luk3:
Hello @petebryan!
I launched black and it modified two lines format.

Could you see if it is already correct?

petebryan (approved):

petebryan:
@lucky-luk3 thats great, I've merged!

Issue #275, opened 2022-01-11
Add support for TIProvider IntSights


FlorianBracq:
Intsights (Rapid7) is a Threat Intel Provider can be integrated with msticpy.

https://intsights.com/
https://intsights.com/resources/investigation-api

Sadly, most of the documentation requires an account, thus cannot be shared here.

petebryan:
@FlorianBracq - great suggestion, this isn't a provider I am familiar with but it would be a great contribution!

PR #276, opened 2022-01-11
Add IntSights support

FlorianBracq:
Simple PR related to #275


petebryan (commented):
Thanks this looks great!
The code looks good, can I just ask that you add a unit test for this with a mocked response. 
Its quite simple to do and you can see examples for the other providers in test_tiproviders.py.

Please let us know if you have any issues with this or want some help.

FlorianBracq:
Hi @petebryan, I've added some mock data based on what I saw in test_tiproviders.py.

Let me know if I missed anything!

ianhelle:
Can you add a section to https://github.com/microsoft/msticpy/blob/main/msticpy/resources/mpconfig_defaults.yaml
so that IntSights shows up in the Config UI?
It should look like the XForce one below. (

```yaml
TIProviders:
  ....
  XForce:
    Args:
      ApiID: *cred_key
      AuthKey: *cred_key
    Primary: bool(default=False)
    Provider: "XForce"
```

FlorianBracq:
@ianhelle: Section added!

petebryan:
So looking at the failing test here there are a couple for minor changes required.

The first is you need to add the IntSight provider to msticpy\tests\testdata\msticpyconfig.yaml so that the TILookup class in the tests correct loads it. 

Second of all in the mocked data you provide for the test you pass in some dates but in the wrong format.
The format you provide in the mocked data is %Y-%m-%d H:%M:%S.%fZ but the IntSights provider requires %Y-%m-%dT%H:%M:%S.%fZ

If you make these changes the test_tiproviders test file should complete properly.

petebryan:
Sorry @FlorianBracq there was another test error that I forgot to include in my last message.

On line 367 of https://github.com/FlorianBracq/msticpy/blob/IntSights/tests/sectools/test_tiproviders.py where you are mocking the response you have the query variable defined as:
 `query = kwargs["params"]["query"]`
 
 However with this provider there is no query element of the params so you should just take the params as is with:
`query = kwargs["params"]`

petebryan (approved):

Issue #246, opened 2021-12-21
Create pinned version of MSTICPy using package hashes for trusted installation


ianhelle:

ianhelle:
We should combine the use of Pipfile/Poetry with pip-compile from pip-tools
https://github.com/jazzband/pip-tools

PR #279, opened 2022-01-12
Added Pipenv Pipfile(.lock) with fixed dependencies

d3vzer0:
Fixed two broken dependencies and converted the requirements.txt to a Pipfile

Two changes to the requirements:
- ipython should be lower than version 8
- renamed precommit (non-existing package) to pre-commit


ianhelle:
Wow! 😃 
We'd had a TODO/Issue about creating a hashed lock file (esp for high sec users) #246. However, I'm not super familiar with how pipenv works.  In a parallel issue I also wanted to shift from setup.py to pyproject.toml - and move all of the extras, etc. into that. #245
I guess pipenv and setup exist in parallel universes. Although it seems like we might have to specify the requirements in both pyproject.toml and the pipfile, which is a bit of a pain. I wonder also if we can generate the lockfile (and maybe the pipfile) from the CI, e.g. for every merge to main.

I'm a bit concerned about the use of a lock file if it pins msticpy to specific dependency versions (which is what I thought lock files did but the lockfile in your PR seems to suggest that multiple versions are ok (e.g. I see multiple hashes for each version >= the install requires version).
In pipfile docs there is this warning about libraries vs applications - https://pipenv.pypa.io/en/latest/advanced/#pipfile-vs-setup-py
Msticpy is definitely the former and we'd break a lot of stuff if we were tied to specific versions.

Assuming that these questions/worries all have good answers, I can add a pre-commit hook to generate the requirements-all.txt from the pipfile.

petebryan:
I've created https://github.com/microsoft/msticpy/pull/278 that handle the IPython issue in the current format.
I am proposing we move to >= 7.23.1 and made some minor code changes.  These changes mean that the code works with 8.0.0 and also removes dependency conflicts.

ianhelle:
We have a pre-commit script that builds requirements-all.txt. I think it would be a good idea to, if the reqs have changed, to generate a pipfile from requirements.txt and re-generate the lock file.
Otherwise we're going to have two separate lists of requirements specs to maintain.
@d3vser - WDYT

Oh also we can't abandon Py 3.6 just yet (I'm hoping a few more weeks) and def can't require Py 3.9. Is this needed for poetry?

d3vzer0:
> We have a pre-commit script that builds requirements-all.txt. I think it would be a good idea to, if the reqs have changed, to generate a pipfile from requirements.txt and re-generate the lock file. Otherwise we're going to have two separate lists of requirements specs to maintain. @d3vser - WDYT
> 
> Oh also we can't abandon Py 3.6 just yet (I'm hoping a few more weeks) and def can't require Py 3.9. Is this needed for poetry?

Sounds like a good idea, I can try to modify to the pre-commit which makes a lock file. Additionally the Pipfile can be modified to require any version of Python 3 instead of 3.9 :)

ianhelle:
The change I made in #287 to the create_reqs_all script (in tools) lets you create a pipfile from the specs in requirements and setup.py (using these args)
```
    parser.add_argument(
        "--pipfile",
        "-i",
        default=False,
        action="store_true",
        required=False,
        help="Generate a Pipfile",
    )
    parser.add_argument(
        "--pyver",
        "-y",
        default="3.8",
        required=False,
        help="Python version to use in the generated Pipfile",
    )
```
Currently it doesn't do this by default in pre-commit but if you think it's a good idea (which I do) we can add this so that there is always a pipfile for anyone who wants to use it.
For the lock file:
1. Should we auto-create this (we'd need to fix to a specific Python version - probably 3.8 for us since we need to support Sentinel AML, which currently only has a Py38 kernel)?
2. We could create instructions in the docs for people to create their own lock file but that seems a bit wonky - i.e. it should be us who are creating the trusted lock file.

Open to any suggestions. Maybe we could switch this PR to add the pipfile (from the script) and a lock file for Py38

PR #293, opened 2022-01-24
SentinelAlert entity creation

petebryan:
Updated Alert entity and created SentinelAlert entity to add new entity capabilities for these data types.
As part of this deprecating existing SecurityAlert entities in nbtools.


msticpy/datamodel/entities/alert.py
```diff
@@ -295,8 +343,11 @@ def _extract_entities(ents: list):
 
 def _find_og_ent(ent, base_ents):
```
ianhelle, line 344: can we rename to _find_original_entity ?
I know this sounds finnicky - I remember reading something about naming vars, etc. so that the name should be as clear as possible about what it is.
Then if you're reviewing code/stack trace you don't need to navigate to the original definition to work out what's going on. It saves a lot of hassle in the long run, even if it seems more tedious when you're coding.
I've tried to change my naming of things since I read this.

msticpy/datamodel/soc/sentinel_alert.py
```diff
@@ -0,0 +1,159 @@
...
+                for item_name, value in ext_ents.items():
+                    self.__dict__[item_name] = value
+            except json.JSONDecodeError:
+                pass
+        if self.AdditionalData:
+            for item_name, value in self.AdditionalData.items():
+                if item_name not in self.__dict__:
+                    self.__dict__[item_name] = value
+
+    def _add_sent_items(self, src_event):
```
ianhelle, line 112: add_sentinel_items?

msticpy/nbtools/security_alert.py
```diff
@@ -9,6 +9,7 @@
 from typing import Dict, Any, List
 
 import pandas as pd
+from deprecated.sphinx import deprecated
```
ianhelle, line 12: Don't you need to import a replacement for SecurityAlert here?
e.g. from ..datamodel.soc import SentinelAlert as SecurityAlert

Ah ok - so SentinelAlert isn't replacing SecurityAlert? Is that too traumatic a shift?

msticpy/nbtools/security_base.py
```diff
@@ -11,6 +11,7 @@
 from typing import Any, Dict, List, Optional, Union
 
 import pandas as pd
+from deprecated.sphinx import deprecated
```
ianhelle, line 14: Would be nice to get rid of this altogether (like SecurityAlert)

msticpy/nbtools/security_event.py
```diff
@@ -7,6 +7,7 @@
 from typing import List, Dict, Any
 
 import pandas as pd
+from deprecated.sphinx import deprecated
```
ianhelle, line 10: Would be nice to get rid of this as well (like SecurityAlert)

msticpy/nbtools/security_alert.py
ianhelle: Maybe we can do that in phase 2

ianhelle (approved):
Left some comments but looks great!

petebryan:
What I would propose is we drop security_alert, security_base, and security_event in 2.0. We will need to update some notebooks before that though as SentinelAlert is not a direct replacement for SecurityAlert and neither is Alert.

PR #300, opened 2022-01-26
Added ability to start Splunk queries as a job/async

d3vzer0:
The default 'oneshot' method is a blocking HTTP call which may often result in a timeout from various proxies. This change runs the query as a 'job' and polls the status until the job is completed (with a progress bar). The results are fetched after the while loop. The original oneshot query can still be used by specifying oneshot=True, ex. splunk_provider.Alerts.list_all_alerts(oneshot=True)


petebryan (approved):

petebryan:
Thanks @d3vzer0 this is another great contribution! The code looks good from a functional standpoint, however it just needs some formatting to meet linting requirements (you should be able to see details in the failed pipeline above but let me know if you have questions).
It also seems like this also breaks our Splunk unit tests so you also need to have a look and update those if possible.

d3vzer0:
@petebryan Got most of the issues resolved, need one addition to fix the unit test :) I'll push the final version tomorrow

petebryan:
Due to an issue we had to reset the main branch so I've recreated your PR here https://github.com/microsoft/msticpy/pull/337

I've also added in elements to the unit tests to process async queries.

PR #303, opened 2022-01-27
Fix QueryTime unit change

kubajir:
We need to make sure we update the actual start/end values when there is a time unit change within the QueryTime widget. Similarly to time range updates. This becomes apparent whenever the unit change isn't accompanied by the default time range change (e.g. days -> weeks).


petebryan (approved):

petebryan:
Hey @kubajir we are putting together a blog on all the contributions we had to MSTICPy in the last month. Are you happy for us to mention you in this blog and if so how do you want us to do this? It could be your name, GitHub username, Twitter handle. Whatever you want really.

kubajir:
Hey @petebryan, sure. Feel free to use my real name (and/or GitHub username), would be nice if you could also include the company name - Chr. Hansen.

PR #308, opened 2022-01-28
Kusto Driver Example Notebooks

liamkirton:
An example of writing custom security data to a Kusto cluster, then doing some basic analysis using the Kusto Driver.

### DIFF
--- a/docs/notebooks/Kusto-Analysis.ipynb
+++ b/docs/notebooks/Kusto-Analysis.ipynb
@@ -26,15 +26,256 @@
    "cell_type": "code",
    "execution_count": 1,
    "metadata": {},
-   "outputs": [],
+   "outputs": [
+    {
+     "data": {
+      "text/html": [
+       "<hr><h4>Starting Notebook initialization...</h4> <br>"
+      ],
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/html": [
+       "msticpy version installed: 1.6.1 latest published: 1.6.1<br>Latest version is installed.<br> <br>"
+      ],
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/html": [
+       "Processing imports.... <br>"
+      ],
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    },
+    {
+     "data": {
+      "text/html": [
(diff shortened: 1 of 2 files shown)


docs/notebooks/Kusto-Analysis.ipynb
```diff
@@ -0,0 +1,1642 @@
+{
```
ianhelle, line 1: <p>You could also just using import msticpy and then msticpy.init_notebook.</p><p>This imports QueryProvider and you can access refresh_config from msticpy.settings.refresh_config()</p><p>This isn't saving you much to be honest but it might lessen the pain one day when we move things around.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/msticpy/pull/308/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='1'/>

ianhelle, line 1: <p>You could also use the MSTICPy b64 decoder - it will (should) recursively decode embedded b64 strings as well.</p><p>The output is a bit janky though. - you'd get 2 rows (one for the next B6 block) decoded_string is the column in the output DF.</p><p>Prob not worth worrying about.</p><p> </p><p>Oddly enough I'm working on little function that isn't yet in msticpy to take obfuscated powershell stuff like that and print it out in a more readable form - not anything to concern you, just an interesting coincidence.</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/msticpy/pull/308/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='27'/>

docs/notebooks/Kusto-Ingest.ipynb
```diff
@@ -0,0 +1,639 @@
+{
```
ianhelle, line 1: <pre class="ql-syntax" spellcheck="false">Line #28.                        # Try and parse datetimes, in various formats, to datetimes
</pre><p>Won't pd.to_datetime() do this for you?</p><p>One thing you might have to do is add utc=True param to force TZ-aware</p>
<br/>

 _Reply via <a href='https://app.reviewnb.com/microsoft/msticpy/pull/308/discussion/'>ReviewNB</a>_ <div id='ReviewNBCommentContext-DoNotDelete' style='display:none' data-state='OPEN' data-cellIndex='5'/>

ianhelle (approved):

ianhelle:
This is cool stuff. Very nice!

docs/notebooks/Kusto-Analysis.ipynb
liamkirton: <p>Sounds sensible. Now I know about init_notebook! :)</p>

liamkirton: <p>Changed. Good to do it the MSTICPy way.</p>

docs/notebooks/Kusto-Ingest.ipynb
liamkirton: That's a better way to achieve the same thing :)

PR #343, opened 2022-03-17
Ianhelle/pivot dataprov selfload 2022 03 15

ianhelle:
Main changes are:

made QueryProviders dynamically add their queries as pivot functions (so you that you can load them at any time without haveing to re-initialize pivot
made query provider names default to modern names (MSSentinel, MSGraph, MDE)
locked loading of queries to match only the folder with the same (lowercase) name as the provider
added mp_pandas_accessors which consolidates a few accessors in a single module.
fixed a couple of circular import issues


ianhelle:
note: the build failure should go away when we merge the other PRs for the V2.0 branch in - I've fixed this bandit issue several times.

PR #351, opened 2022-03-30
Pebryan/2022 3 29 auth updates

petebryan:
New features:
  - Added new feature to allow for authentication to an Azure AD app via delegated logon.
  - Implemented this capability in the Security Graph Driver

Updates:
 - Used AzureDefaultCredentials in Azure auth scenarios to replace ChainedCredential
 - Added support for KQL auth against other tenants


msticpy/config/ce_azure.py
```diff
@@ -37,6 +37,9 @@ class CEAzure(CESimpleSettings):
     <li>msi - use Managed Service Identity credentials, if available</li>
     <li>cli - use Azure CLI credentials, if available</li>
     <li>interactive - use interactive browser authentication</li>
+    <li>vscode" - to use VSCode credentials</li>
```
ianhelle, line 40: I think we need to add these options to the config yaml right?
in resources\mpconfig_defaults.yaml
```yaml
Azure:
  cloud: enum(required=False, options=[global; cn; usgov; de], default=global)
  auth_methods: m_enum(required=False, options=[env; msi; cli; interactive; vscode; powershell; cache], default=["cli"; "msi", "interactive"])
```
Not sure if we want to change the defaults.

msticpy/common/msal_auth.py
```diff
@@ -0,0 +1,138 @@
+# -------------------------------------------------------------------------
```
ianhelle, line 1: This is very cool!!!!

msticpy/common/azure_auth_core.py
```diff
@@ -205,16 +205,30 @@ def _az_connect_core(
...
-        raise MsticpyAzureConnectionError(
-            "Unknown authentication option, valid options are; env, cli, msi, interactive"
-        ) from err
+    if auth_methods:
+        for method in auth_methods:
+            if method in _AUTH_MAPPINGS:
+                _AUTH_MAPPINGS[method] = False
+        creds = DefaultAzureCredential(
+            authority=aad_uri,
+            exclude_cli_credential=_AUTH_MAPPINGS["cli"],
```
ianhelle, line 217: Is this logic the wrong way around?
Ah no I guess not. Maybe we should rename _AUTH_MAPPINGS to something like _EXCLUDED_AUTH?
Then at least setting something in there to True makes a bit more sense.

msticpy/data/drivers/odata_driver.py
```diff
@@ -208,6 +247,18 @@ def query_with_results(self, query: str, **kwargs) -> Tuple[pd.DataFrame, Any]:
             req_url = self.request_uri + query  # type: ignore
             response = httpx.get(url=req_url, headers=self.req_headers)
 
+        # If it fails when using MSAL creds try getting a new one token
```
ianhelle, line 250: Is this meant to be commented out?
Maybe delete, if it's not being used.

msticpy/data/drivers/security_graph_driver.py
```diff
@@ -23,27 +24,37 @@ class SecurityGraphDriver(OData):
...
             "client_secret": None,
             "grant_type": "client_credentials",
             "scope": f"{azure_cloud.endpoints.microsoft_graph_resource_id}/.default",
         }
-        self.oauth_url = (
-            f"{azure_cloud.endpoints.active_directory}/" "{tenantId}/oauth2/v2.0/token"
-        )
+
+        if delegated_auth:
+            self.oauth_url = f"{azure_cloud.endpoints.active_directory}/" "{tenantId}"
```
ianhelle, line 52: just join into a single f-string?

ianhelle (approved):
This is great stuff! Thanks for digging into this.
I've left a couple of comments. The most significant one is the config YAML stuff.
Nothing very onerous though.
Any unit tests?

petebryan:
@ianhelle - added unit tests and some minor fixes for your approval!

msticpy/common/azure_auth_core.py
```diff
@@ -320,22 +333,6 @@ def _filter_all_warnings(record) -> bool:
     return True
 
 
-def _create_auth_options(cloud: str = None, tenant_id: str = None) -> dict:
```
ianhelle, line 323: In the mp_config_load branch I've made some changes to this section to fix a bug. 
You can overwrite this section completely when you merge.

ianhelle (approved):

PR #358, opened 2022-04-12
Some fixes to Kusto common_imports

ianhelle:
- now works with Kusto config entry without instance suffix
- can now supply cluster ALIAS (instance name) instead of actual cluster name in connect or query
- added explicit "database" key in query files - can be used instead of the more opaque "data_family.database" encoding
  in the data_famiies key.
Fixed documentation in DataProv-Kusto.rst to correct inaccuracies and update sections on query templates and configuration


msticpy/data/drivers/kusto_driver.py
```diff
@@ -240,10 +247,20 @@ def _lookup_cluster(self, cluster: str):
...
-        return None
+        return next(
+            (
+                kusto_config["cluster"]
+                for cluster_key, kusto_config in self._kusto_settings.items()
+                if (
+                    cluster_key.startswith(f"https://{cluster.casefold()}.")
+                    or (
+                        kusto_config.get("alias", "").casefold()  # type: ignore
+                        == cluster.casefold()
```
petebryan, line 258: This is a great addition

petebryan (approved):

Issue #366, opened 2022-04-13
MicrosoftSentinel constructor demands default workspace in config file


aeetos:
A `MicrosoftSentinel` class constructed without parameters does not work properly without a "Default" section in the configuration file. The error message is somewhat unhelpful and the self-correction behavior rather unsettling. 

**To repro:**
1. Ensure no `config.json` file is present in the current directory
2. Ensure a `msticpyconfig.yaml` file exists, is formatted properly, has at least one Azure workspace configured, but without a default workspace. For example:
```
AzureSentinel:
  Workspaces:
    TestWorkspace:
      ResourceGroup: rgtest
      SubscriptionId: 3f3fb3df-3c79-4407-8043-4a765430db0e
      TenantId: 3f3fb3df-3c79-4407-8043-4a765430db0e
      WorkspaceId: 3f3fb3df-3c79-4407-8043-4a765430db0e
```
3. Init the notebook, ensure no warnings are generated about the config file:
```
import msticpy
msticpy.init_notebook(namespace=globals())
```
4. Import the class and attempt to instantiate the object:
```
sentinel = MicrosoftSentinel()
```
5. Possibly wait a few seconds, then observe the warning output:
```
Warning
Could not find Microsoft Sentinel settings in msticpyconfig.yaml or a config.json 
(in the current directory or via a MSTICPYCONFIG variable.)
We found the file '../msticpy/dev/config.json' and will use this.
We recommend using an explicit msticpyconfig.yaml specified using the
MSTICPYCONFIG environment variable. See:
https://msticpy.readthedocs.io/en/latest/data_acquisition/DataProviders.html#connecting-to-an-azure-sentinel-workspace
and
https://msticpy.readthedocs.io/en/latest/getting_started/msticpyconfig.html
for more details.
Also see the notebooks:
https://github.com/Azure/Azure-Sentinel-Notebooks/blob/master/ConfiguringNotebookEnvironment.ipynb
https://github.com/Azure/Azure-Sentinel-Notebooks/blob/master/A%20Getting%20Started%20Guide%20For%20Azure%20Sentinel%20ML%20Notebooks.ipynb
```

The code works as expected if a default section exists (although a warning is raised if the WorkspaceName setting is not present, issue #365 should help fix that).

**Expected behavior**
1. If the `msticpyconfig.yaml` exists and a default section exists in the file, use those settings
2. If no default section exists, and only one workspace config is present, use that by default
[...]

Issue #365, opened 2022-04-13
Store Sentinel workspace name in config file


aeetos:
Some interactions with the Microsoft Sentinel API require a workspace name, not just a workspace ID. 

We should store the workspace name in the configuration file (and provide a mechanism to edit it interactively) in addition to the workspace ID. We may also need an easy way of translating between the two internally (so the user can specify either).

This is especially relevant for the `MicrosoftSentinel` objects used to interact with the Sentinel API for important use cases including writing threat intelligence indicators to watchlists. The constructor for that object, and presumably also the API, requires a workspace name, not a workspace ID.

petebryan:
So the ability to add and use WorkspaceName in the config file and for this to be used by MicrosoftSentinel is already there. What we don't currently have is the ability to add this to the config using the config widget - we should add that.

PR #376, opened 2022-04-22
Added new Sentinel Search Features:

petebryan:
Create a new search
Check the status of a search
Delete a completed search
Includes docs and unit tests

Updated WorkspaceConfig:
If one workspace in config but not called default
it is still used by default

Addresses elements of #366 and #365


ianhelle (approved):

ianhelle:
There is a code scanning error on the tests - about a regex with unescaped "." - otherwise looks good

PR #384, opened 2022-05-04
Ianhelle/folium accessor 2022 04 30

ianhelle:


petebryan (approved):
This is great, so much nicer to use!

PR #409, opened 2022-05-19
Adding full Delegated Auth support to all OData Drivers

petebryan:
Updated OData drivers to allow for Delegated auth settings to be passed when connecting.
Includes the ability to use Delegated Auth as well as the method. 
Added documentation on how to use the feature.

PR #410, opened 2022-05-19
Adding OData Delegated Auth Support into 2.0

petebryan:
Adding Odata delegated auth items from #409 into v2.0.0


ianhelle (approved):
Cool - the comments from the main review would be the same here.

PR #417, opened 2022-05-25
Fix for list_hunting_queries function

pensivepaddle:
Fix for list_hunting_queries function, referred to alert_rules api, which does not contain such.
Rather going towards savedSearches endpoint. 
It could also be pointed out somewhere that this is for custom queries only, i.e. Provider="Custom Queries"


petebryan (approved):

petebryan:
Thanks @pensivepaddle - this API is returning some of the built in hunting queries but not all of them. I will raise this with the product team and see if we can identify the reason for this discrepency.

PR #456, opened 2022-06-20
Added features to support new notebooklet development.

petebryan:
- Fixed issue with KQL Driver passing UA
- Added ability for cross provider query sharing
- Added timeout feature to screenshot
- Fixed issues with Alert Selector
- Added new queries for Windows Events and Syslog


msticpy/data/queries/m365d/kql_mdatp_process.yaml
```diff
@@ -22,15 +22,19 @@ defaults:
...
 sources:
   list_host_processes:
     description: Lists all process creations for a host
     metadata:
     args:
       query: '
         {table}
-        | where Timestamp >= datetime({start})
-        | where Timestamp <= datetime({end})
+        | where {time_column} >= datetime({start})
```
ianhelle, line 36: time_column seems a bit of weird name - maybe time_stamp?

ianhelle (dismissed):

msticpy/data/queries/m365d/kql_mdatp_process.yaml
petebryan: So I picked this as we use time column (or time_col) everywhere else we parameterize the column name such as in the plots, and so thought it would keep things consistent.

ianhelle (approved):

PR #459, opened 2022-07-01
Adding Malware Bazaar module as TI provider

fr0gger:
I coded a module for requesting Malware Bazaar API. The module can be used to request specific IOCs such as:

- 'hash': the sha256 hash of your sample (nb: the module does not calculate the hash automatically)
- 'tag': the tag used on Malware Bazaar to retrieve a set of specific sample. You can use the 'limit' (default is 50)
- 'filetype': the type of files you want to retrieve. Limit is 50 by default
- 'clamav': the Clamav Signature that matches the samples you want to retrieve.
- 'imphash': the imphash of files you want to retrieve.
- 'dhash': the icon hash that matches the samples you want to retrieve.
- 'yara': the Yara rule that matches the samples. 
- 'tlsh': the tlsh that matches the samples.
- 'telfhash': the Telfhash that matches the samples.
- 'issuerinfo': the certificate issuer that is used in the matching samples. 
- 'subjectinfo': the certificate subject that used by the samples. 
- 'certifcate': the serial number of the certificate.

It can also be used to download a sample, get the latest samples uploaded to MB or retrieve the CSCB. 

There is also a notebook that demonstrates how to use the module. :)


msticpy/context/tiproviders/mblookup.py
```diff
@@ -0,0 +1,317 @@
...
+    - Download malware sample from MalwareBazaar
+    - Query Code Signing Certificate Blocklist (CSCB)
+    - Get the recent addition from MalwareBazaar
+
+"""
+from enum import Enum
+from typing import Set
+
+import json
+import requests
```
petebryan, line 24: We've tried to standardize on using httpx for web requests - are you able to use that package here instead?

msticpy/context/tiproviders/mblookup.py
```diff
@@ -0,0 +1,317 @@
...
+        MBEntityType.TELFHASH,
+        MBEntityType.CODESIGNISSUER,
+        MBEntityType.CODESIGNSUBJECT,
+        MBEntityType.CODESIGNSN,
+        MBEntityType.GIMPHASH,
+    }
+
+    def __init__(self, mb_key=None):
+        """Init function to get the API key if necessary"""
+        self.mb_key = mb_key # or_get_mb_api_key()
```
petebryan, line 86: is the or statement meant to be commented out?

msticpy/context/tiproviders/mblookup.py
```diff
@@ -0,0 +1,317 @@
...
+
+    def lookup_ioc(self, observable: str, mb_type: str, limit=10) -> pd.DataFrame:
+        """
+        Lookup an IOC in MalwareBazaar.
```
petebryan, line 90: Can we ask that you expand out the docstrings with Parameters and Returns sections please?

msticpy/context/tiproviders/mblookup.py
```diff
@@ -0,0 +1,317 @@
...
+        """
+
+        if MBEntityType(mb_type) not in self._SUPPORTED_MB_TYPES:
+            raise KeyError(
+                f"Property type {mb_type} not supported",
+                "Valid types are",
+                ", ".join(x.value for x in MBEntityType.__members__.values()),
+            )
+
+        if mb_type == "hash":
```
petebryan, line 100: There is a lot of very similar code in this if chain.
I wonder if we could simplify it with a generic function that gets passed a data parameter depending on the type of IoC.
Something like:


```
def _make_mb_request(data):
    res = httpx.post(_BASE_URL, data=data)
    res_data = json.loads(res.text) # or if it works ok res.json()
    if res["query_status"] == 'ok':
        return json_normalize(res['data'])
    return res["query_status"]

if mb_type == "hash":
    return _make_mb_request({'query': 'get_info', "hash": observable})
```

msticpy/context/tiproviders/mblookup.py
```diff
@@ -0,0 +1,317 @@
...
+            data = {'query': 'get_file', "sha256_hash": sha2}
+            res = requests.post(_BASE_URL, data=data)
+            return res.content
+        except requests.exceptions.RequestException as err:
+            return err
+
+    def get_recent(self, selector):
+        """ Get the recent MB additions.
+
+            selector:
```
petebryan, line 283: Need some type hints here, plus can you please use numpy docstring format - [numpy docstring guide](https://numpydoc.readthedocs.io/en/latest/format.html)

petebryan (changes requested):

petebryan:
Thanks for this @fr0gger - a great addition!

I've added a few comments - nothing fundamental just a few suggestions to keep it aligned with some of our other code, and to reduce a bit of code reuse (that's all Ian's influence on me!). Feel free to hit me up on Teams if you want to chat more about anything.

msticpy/context/tiproviders/mblookup.py
fr0gger: Perfect I'll do that asap

msticpy/context/tiproviders/mblookup.py
fr0gger: Actually, for the current version there is no need to have the API I just let this piece of code for further development

msticpy/context/tiproviders/mblookup.py
fr0gger: Sure

msticpy/context/tiproviders/mblookup.py
fr0gger: Sounds good I can do that

fr0gger:
Hey @petebryan, updated the module with your suggestions. Let me know if further updates are needed. :)

petebryan (approved):

petebryan:
Just made some minor changes to make the linters happy

PR #481, opened 2022-08-07
Sumologic driver: custom dtypes options+fix, add paging, remove days duration int casting

juju4:
* add optional datetime/numeric typing and fix native metadata fields
  * time_columns: for extra datetime columns
  * numeric_columns: for extra int columns
(default panda column is object/string)
* add paging results else limit of 10k 
* remove int casting preventing to use partial day/float 

Add example queries in documentation notebook for easy testing


msticpy/data/drivers/sumologic_driver.py
```diff
@@ -287,10 +291,12 @@ def _poll_job_status(self, searchjob, verbosity):
 
     # pylint: disable=inconsistent-return-statements
     # I don't think there are any - everything returns a list
-    def _get_job_results(self, searchjob, status, qry_count, force_mssg_rstls, limit):
+    def _get_job_results(
```
ianhelle, line 294: This is a scarily long function. Is it possible to break it up into subfunctions that each handled the 4 cases:
- get_results_with_limit_force()
- get_results_no_limit_force()
- get_results_with_limit_gt_count()
- get_results_with_limit_lte_count()
- get_results_no_limit_gt_count()
- get_results_no_limit_lte_count()

Or something along those lines.
It would be easier to see the logic of how each option is selected and would abstract this from the mechanics of setting up the service query.

msticpy/data/drivers/sumologic_driver.py
```diff
@@ -425,9 +527,31 @@ def query(
...
         for col in dataframe_res.columns:
-            if col in ("map._count", "map._timeslice"):
-                dataframe_res[col] = pd.to_numeric(dataframe_res[col])
+            try:
+                if (
+                    col
+                    in ["map._count", "map._collectorid", "map._messageid", "map._size"]
+                    + numeric_columns
+                ):
+                    dataframe_res[col] = pd.to_numeric(dataframe_res[col])
```
ianhelle, line 540: Apparently dynamically adding pandas columns like this can result in DF fragmentation, which results in warnings from pandas (and presumably poorer perf).
One option would be to return a df.copy() at the end of the function.

ianhelle (changes requested):

msticpy/data/drivers/sumologic_driver.py
juju4: done. that was already on my list. just ongoing work. split in 2 only is enough for linting to be happy.

msticpy/data/drivers/sumologic_driver.py
juju4: changed

ianhelle (approved):

juju4:
Thanks @ianhelle !

PR #491, opened 2022-08-18
Write Sentinel queries to YAML for Github Browser

jannieli:
Basic code to load in Sentinel queries from Github and write out to YAML files to be read in by Github Browser


msticpy/data/drivers/github_data_types.py
```diff
@@ -0,0 +1,288 @@
...
+    # Download the Microsoft Sentinel Github repo as ZIP
+    azsentinel_git_url = 'https://github.com/Azure/Azure-Sentinel/archive/master.zip'
+    get_sentinel_queries_from_github(git_url=azsentinel_git_url, outputdir=def_path)
+    base_dir = str(def_path) + "/Azure-Sentinel-master"
+    
+    print('processing queries')
+    detection_queries = parse_yaml(parent_dir=base_dir, child_dir="Detections")
+    hunting_queries = parse_yaml(parent_dir=base_dir, child_dir="Hunting Queries")
+
+    d_child_dir = "Detections"
```
ianhelle, line 277: Shouldn't need these sections I think if you're extracting and saving the folder name when reading the yaml files.

msticpy/data/drivers/github_data_types.py
```diff
@@ -0,0 +1,288 @@
...
+    outputdir : str, optional
+        Provide absolute path to the output folder to save downloaded archive (e.g. '/usr/home' or 'C:\downloads'),
+        If no path provided, it will download to .msticpy dir under Azure-Sentinel directory.
+    """
+
+    if outputdir is None:
+        outputdir = Path.joinpath(Path("~").expanduser(), ".msticpy", "Azure-Sentinel")
+
+    try:
+        with requests.get(git_url, stream=True) as response:
```
ianhelle, line 75: We've stopped using requests - use httpx instead - it's a drop-in replacement.
```
from ..common.pkg_config import get_http_timeout
from ..common.utility import mp_ua_header

....

response = httpx.get(
            submit_url, timeout=get_http_timeout(), headers=mp_ua_header()
        )
```

ianhelle (changes requested):
I've added quite a lot of comments - but these are intended as educational (indoctrinating you in the msticpy culture :-))
Some things that might help - we use a lot of tools that will flag a bunch of things like formatting and more efficient ways of doing things - and also warn you if a function is too complex. 
I often do my early dev work in Jupyter and then paste into VScode to do final editing.
If you follow the stuff I sent you earlier about installing the tools (cd msticpy, pip install --r requirements-dev.txt). VScode Python extension will discover these and start using them.
Then you'll get interactive squigglies and warnings about things the tools flag.
It's a bit of a pain at first but you get used to it.

msticpy/data/drivers/github_data_types.py
```diff
@@ -0,0 +1,288 @@
...
+from pandas import json_normalize
+from tqdm.notebook import tqdm
+from datetime import date
+from datetime import datetime
+from requests_html import HTMLSession
+from collections import Counter
+
+@attrs
+class Query(object):
+    id: str = attrib(factory=str)
```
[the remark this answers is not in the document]
jannieli: Seems if you call Query.name or something left empty without a "=" - it gives an AttributeError whereas with it, it'll just return an empty string. Would that be preferable to an error?

msticpy/data/drivers/github_data_types.py
```diff
@@ -0,0 +1,288 @@
...
+    yaml_queries = glob.glob(f"{parent_dir}/{child_dir}/**/*.yaml", recursive=True)
+    no_qf = 0
+    total_ct = 0
+    query_attrs = []
+    for query in yaml_queries: 
+        with open(query, "r", encoding="utf-8", errors="ignore") as f:
+#             parsed_yaml_df = json_normalize(yaml.load(f, Loader=yaml.FullLoader))
+            parsed_yaml_dict = yaml.load(f, Loader=yaml.FullLoader)
+
+            if 'relevantTechniques' in parsed_yaml_dict.keys():
```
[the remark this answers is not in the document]
jannieli: Tried this and it works - but wondering since when I used the dict like dict[key], empty fields or when there were no matches threw an error but this does not - do we want to be more aware of missing things or I guess if it returns something empty or quietly fails it'll just not exist in the data. Thoughts?

msticpy/data/drivers/github_data_types.py
jannieli: httpx seems to get the state of the web page rather than downloading it - may be wrong and will look into this and update soon!

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+    Returns
+    -------
+    bool
+        True if succeeded; False if an error occurred
+
+    """
+    query_dict_by_folder = _organize_query_list_by_folder(query_list)
+    all_folders = query_dict_by_folder.keys()
+
+    for source_folder in all_folders:
```
ianhelle, line 331: Might be good to have tqdm on this loop since I would guess it's quite time-consuming.

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+
+    Returns
+    -------
+    SentinelQuery
+        Returns an attrs object called SentinelQuery with all the YAML query information
+
+    """
+    try:
+        parsed_yaml_dict = yaml.load(yaml_text, Loader=yaml.SafeLoader)
+        new_query = SentinelQuery(
```
ianhelle, line 196: Just a thought here.
You could implement the SentinelQuery using exactly the same names as the source data (most already are)
Set defaults on the ones that need defaults ("",, Factory(list), Factory(dict), etc.
Then you can just call it like this
```
    new_query = SentinelQuery(**parsed_yaml_dict)
```

You can add additional params as keywords
```
    new_query = SentinelQuery(**parsed_yaml_dict, source_file=yaml_path, query_type=query_type)
```

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+            tactics=parsed_yaml_dict.get("tactics", ""),
+            relevant_techniques=parsed_yaml_dict.get("relevantTechniques", []),
+            query=parsed_yaml_dict.get("query", ""),
+            version=parsed_yaml_dict.get("version", ""),
+            kind=parsed_yaml_dict.get("kind", ""),
+            folder_name=yaml_path.replace("\\", "/").split("/")[-2],
+            source_file_name=yaml_path,
+            query_type=query_type,
+        )
+        if new_query is None:
```
ianhelle, line 216: How would new_query ever be None?

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+    query_type : str
+        Either "Hunting Queries" or "Detections" or otherwise named query category
+
+    Returns
+    -------
+    SentinelQuery
+        Returns an attrs object called SentinelQuery with all the YAML query information
+
+    """
+    try:
```
ianhelle, line 194: I'd have separate try block around load yaml and catch yaml.error.YamlError

Then a separate one around the creation of SentinelQuery that catches (TypeError, AttributeError)

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+            source_file_name=yaml_path,
+            query_type=query_type,
+        )
+        if new_query is None:
+            print("No info for query - query is None")
+            print(yaml_path)
+            print(yaml_text)
+        return new_query
+
+    except Exception as error:
```
ianhelle, line 222: In the except blocks for the previous section, would be good places to put
logger.warning("Failed to parse yaml for %s", yaml_path)

logger.warning("Format error when parsing %s", yaml_path)

tests/data/drivers/test_sentinel_query_reader.py
```diff
@@ -0,0 +1,214 @@
...
+def test_get_sentinel_queries_from_github():
+    assert get_sentinel_queries_from_github(outputdir=os.getcwd())
+
+
+def test_read_yaml_files():
+    yaml_files = read_yaml_files(parent_dir=BASE_DIR_TEST_FOLDER, child_dir="Detections")
+    assert yaml_files[str(BASE_DIR_TEST_FOLDER.joinpath("Detections/Anomalies/UnusualAnomaly.yaml"))]
+
+
+def test__import_sentinel_query():
```
ianhelle, line 55: extra "_" in name

tests/data/drivers/test_sentinel_query_reader.py
```diff
@@ -0,0 +1,214 @@
...
+@patch(f"{_SENTINEL_QUERY_READER}.get_sentinel_queries_from_github")
+def test_get_sentinel_queries_from_github(get_sentinel_queries_from_github):
+    get_sentinel_queries_from_github.return_value = True
+    assert get_sentinel_queries_from_github(outputdir=os.getcwd())
+    get_sentinel_queries_from_github.assert_called_once()
+
+
+# original test case for downloading github zip
+@pytest.mark.skip(reason="requires downloading the file directly during the test")
+def test_get_sentinel_queries_from_github():
```
ianhelle, line 46: You could, if you wanted, create a unit test patch that returned a zip file - emulating your function.
Maybe we don't need this but it's a possibility.

ianhelle (changes requested):
Let me know if any of the comments are mysterious.
I can def help with the logging - or actually any of this. Would be fantastic to get this in.

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+    -------
+    dict
+        Returns a dictionary mapping each folder name with the SentinelQuery objects associated
+        with it
+
+    """
+    queries_by_folder = {}
+    for query in query_list:
+        if query.folder_name == "":
+            print(query)
```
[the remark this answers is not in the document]
jannieli: I don't think it's likely but it has happened a couple times during testing so I think that's when I added the condition.

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -0,0 +1,452 @@
...
+    custom_details: dict = attrib(factory=dict)
+    alert_details_override: dict = attrib(factory=dict)
+    version: str = attrib(factory=str)
+    kind: str = attrib(factory=str)
+    folder_name: str = attrib(factory=str)
+    source_file_name: str = attrib(factory=str)
+    query_type: str = attrib(factory=str)
+
+
+def get_sentinel_queries_from_github(
```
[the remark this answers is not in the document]
jannieli: Updated the function that currently creates the YAML files given a downloaded folder to include the download portion

msticpy/data/drivers/sentinel_query_reader.py
jannieli: For some reason writing out the files actually takes < 2 seconds, but I'll log out the folders being written instead of printing so that the general function should be able to see those too.

msticpy/data/drivers/sentinel_query_reader.py
jannieli: I didn't use the exact same names because the camel case didn't match what the Python linters were looking for. Totally happy to change this back to the original source name but keeping it for now - let me know if it would be better to switch over!

msticpy/data/drivers/sentinel_query_reader.py
jannieli: I think this was something left over from before- got rid of it!

msticpy/data/drivers/sentinel_query_reader.py
jannieli: Realized that since I'm using "get" to get items from parsed_yaml_dict, it doesn't have errors if the key doesn't exist so just wrapping the YAML for now.

tests/data/drivers/test_sentinel_query_reader.py
jannieli: There are two test functions - one for _import_sentinel_query and one for import_sentinel_query so that's why there's an extra "_"

tests/data/drivers/test_sentinel_query_reader.py
jannieli: Will look into this!

ianhelle (approved):

PR #496, opened 2022-09-07
Resourcegraph - Incomplete list returned

pensivepaddle:
Perhaps not the best solution, but it works - running the ResourceGraph driver for "PolicyResources" without the addition returns incomplete list of results, with no obvious reason. 

Noticed in the portal that not all results for a "PolicyResources"-query contain subscriptionId value, and the missing results correspond to those with "blank" values in the column.

By adding a blank string to the list, the expected results were returned.


msticpy/data/drivers/resource_graph_driver.py
```diff
@@ -93,7 +93,7 @@ def connect(self, connection_str: str = None, **kwargs):
         )
         self.subscription_ids = [
             sub.subscription_id
-            for sub in self.sub_client.subscriptions.list()  # type: ignore
+            for sub in self.sub_client.subscriptions.list().append("")  # type: ignore
```
ianhelle, line 96: I think this should be something like:
```python
self.subscription_ids = [
            sub.subscription_id
            for sub in self.sub_client.subscriptions.list()
]
self.subscription_ids.append("")
```
.append() adds an item to an existing list but doesn't return anything.
Also "" is not going to have a "subscription_id" attribute.

ianhelle (changes requested):

ianhelle (approved):

ianhelle (approved):

PR #499, opened 2022-09-10
Fix breaking issues in Auth and Browshot

petebryan:
Browshot:
 - Missing f string resulted in incorrect building of URL which mean created screenshots could not be retreived.

Azure Auth Core:
 - Filtering of None event when env client creation script broke creation of chained credentials in certain circumstances. Removed filter that removed None objects.


msticpy/auth/azure_auth_core.py
```diff
@@ -302,8 +302,7 @@ def _build_chained_creds(
         raise MsticpyAzureConfigError(
             "At least one valid authentication method required."
         )
-
-    return ChainedTokenCredential(*[client for client in clients if client])
+    return ChainedTokenCredential(*clients)
```
ianhelle, line 305: I know my change didn't look like it did anything but the env client returns None if there are no env vars set 
- so clients is like [X, Y, None, Z] (see line #65)
If this is on the list of auth methods, and gets exercised it throws an exception if the vars are not set.
I'm assuming this caused a problem though. 
Maybe change "if client" to "if client is not None"?

ianhelle (approved):
Looks good although we ought to come to a resolution on the env client auth stuff.

msticpy/auth/azure_auth_core.py
petebryan: So yeah this causes the issue - if you call this function with just the Environment Variable as your only auth option but without any env variables set clients is []. This results in ChainedTokenCredential throwing an exception.  As this is something KeyvaultClient does every time you run init_notebook() you get an exception unless you have env variables set for auth.
This solution means that clients is [None] which ChainedTokenCredential is fine with - it obviously can't auth with it but it doesn't create any issues.

petebryan: ChainedTokenCredential needs something passing to it - otherwise it throws an exception - this is happening whenever you call az_connect_core(auth_methods=["env"]) when you don't have environment variables set.
This issue is KeyVaultClient does just this be default and so it and everything that uses it like init_notebook() throws exceptions.
I've made a change so we still filter out None clients but we will at least pass an empty list. This seems to solve the issue and hopefully meets your needs.

ianhelle (approved):

PR #504, opened 2022-09-14
Rolling back change on _build_chained_creds

FlorianBracq:
ChainedTokenCredential constructor expects a TokenCredential but is sent a list https://docs.microsoft.com/en-us/python/api/azure-identity/azure.identity.chainedtokencredential?view=azure-python

Error triggered when calling `self._get_working_kv_client(kv_client)` (l181 of msticpy/auth/keyvault_client.py)

`azure.core.exceptions.ClientAuthenticationError: ChainedTokenCredential failed to retrieve a token from the included credentials.
Attempted credentials:
	list: 'list' object has no attribute 'get_token'
To mitigate this issue, please refer to the troubleshooting guidelines here at https://aka.ms/azsdk/python/identity/defaultazurecredential/troubleshoot.`

### DIFF
--- a/msticpy/auth/azure_auth_core.py
+++ b/msticpy/auth/azure_auth_core.py
@@ -302,7 +302,7 @@ def _build_chained_creds(
         raise MsticpyAzureConfigError(
             "At least one valid authentication method required."
         )
-    return ChainedTokenCredential([client for client in clients if client])  # type: ignore
+    return ChainedTokenCredential(*[client for client in clients if client])  # type: ignore
 
 
 class _AzCachedConnect:


ianhelle:
Pete and I just found this as well. Thx so much for the fix.
This was a very last-minute change that we made - good lessons for the future when we are futzing with such core code.

I've also got a PR - turns out that leaving None in the list (an edge case where client requests environment credential type but doesn't have env vars set) let's it work - or at least produces nicer errors - when this happens.
- simplified to ChainedTokenCredential(*clients)

FlorianBracq:
Fixed in #505

PR #532, opened 2022-10-01
Added Sentinel TI integration features.

petebryan:
### New Sentinel Threat Intelligence Integration Features

We have expanded the Microsoft Sentinel API integrations to provide a method to integrate with the Threat Intelligence feature.
This provides users with the ability to:

- Get a list of all indicators in a Workspace
- Get details of a specific indicator
- Query for indicators by key values
- Create indicators, individually and in bulk
- Update indicators
- Delete indicators

These features integrate with the existing Microsoft Sentinel model.

**Getting Indicators**

Getting all indicators is as simple as:
```
import msticpy as mp
sent = mp.MicrosoftSentinel()
sent. Connect()

sent.get_all_indicators()
```

image

**Updating Indicators**

Indicators can be updated using `update_indicator` and passing in the incident ID to update:

```
sent.update_indicator(indicator_id="89805941-6adf-424a-b0f6-32bb5e93bf35", confidence=90)
```

Full details of these features can be found in the [MSTICPy documentation](https://msticpy.readthedocs.io/en/latest/data_acquisition/Sentinel.html).


docs/source/data_acquisition/SentinelTI.rst
```diff
@@ -0,0 +1,149 @@
...
+https://learn.microsoft.com/azure/sentinel/work-with-threat-indicators
+
+Get Indicators
+--------------
+
+It is possible to return details of all indicators within a workspace.
+Whilst it is possible to access these indicator details via the indicator table in the Workspace, you can also interact
+with them via the Microsoft Sentinel APIs which are utilized in these functions.
+
+See :py:meth:`get_all_indicators <msticpy.context.azure_sentinel.MicrosoftSentinel.get_all_indicators>`
```
ianhelle, line 15: These paths look wrong. Shouldn't it be ...context.azure.sentinel_core.MicrosoftSentinel ?

docs/source/data_acquisition/SentinelTI.rst
```diff
@@ -0,0 +1,149 @@
...
+
+.. code:: ipython3
+
+    azs.get_all_indicators()
```
ianhelle, line 19: maybe sentinel.get_all_indicators (vs "azs")?

msticpy/context/azure/sentinel_ti.py
```diff
@@ -0,0 +1,455 @@
...
+
+    def create_indicator(
+        self,
+        indicator: str,
+        ioc_type: str,
+        name: str = "TI Indicator",
+        confidence: int = 0,
+        silent: bool = False,
+        **kwargs,
+    ):
```
ianhelle, line 89: Missing return type annotation and doc return description

msticpy/context/azure/sentinel_ti.py
```diff
@@ -0,0 +1,455 @@
...
+        """
+        Create a new indicator within the Microsoft Sentinel workspace.
+
+        Parameters
+        ----------
+        indicator : str
+            The indicator to create - i.e. IP address, domain name etc.
+        ioc_type : str
```
ianhelle, line 97: Wondering if we should allow the the IoC types that we use as aliases?
dns, file_hash, ipv4, ipv6?

This is the ioc enum from iocextract.py
```python
    ipv4 = "ipv4"
    ipv6 = "ipv6"
    dns = "dns"
    url = "url"
    md5_hash = "md5_hash"
    sha1_hash = "sha1_hash"
    sha256_hash = "sha256_hash"
    file_hash = "file_hash"
    email = "email"
    windows_path = "windows_path"
    linux_path = "linux_path"
    hostname = "hostname"
```

Also, this list doesn't match the list you check against in the body of the function.
(if we allowed file_hash, we'd have to check length 32 (md5), 40 (sha1), 64 (sha256) and map it to the appropriate type)

msticpy/context/azure/sentinel_ti.py
```diff
@@ -0,0 +1,455 @@
...
+        Bulk create indicators from a DataFrame.
+
+        Parameters
+        ----------
+        data : pd.DataFrame
+            A dataframe containing indicators and indicator types
+        indicator_column : str, optional
+            The column containing indicator values to create, by default "Observable"
+        indicator_type_column : str, optional
+            The column containing indicator type values, by default "IoCType"
```
ianhelle, line 197: Might want to add the acceptable types here.
Mapping from Msticpy types here might be messier. I guess we could create a function that you could use in an apply or maybe assign.
```python
_SENTINEL_IOC_TYPES = {
    "ipv4": "ipaddr_v4",
    # etc.
}

def map_to_sentinel_ioc_type(ioc_type: str):
    if ioc_type in _SENTINEL_IOC_TYPES.values():
        return ioc_type
    return _SENTINEL_IOC_TYPES.get(ioc_type)
```
or something like that

msticpy/context/azure/sentinel_ti.py
```diff
@@ -0,0 +1,455 @@
...
+            ti_url,
+            headers=get_api_headers(self.token),  # type: ignore
+            params=params,
+            timeout=get_http_timeout(),
+        )
+        if response.status_code not in (200, 204):
+            raise CloudError(response=response)
+        print("Indicator deleted.")
+
+    def query_indicators(self, **kwargs) -> pd.DataFrame:
```
ianhelle, line 355: Seems odd that you cannot query for an indicator value (i.e. like aa IP address)

msticpy/transform/iocextract.py
```diff
@@ -125,18 +127,35 @@ class IoCExtract:
...
-            (?P<protocol>(https?|ftp|telnet|ldap|file)://)
+            (?P<protocol>(https?|s?ftps?|telnet|ldap|file)://)
             (?P<userinfo>([a-z0-9-._~!$&\'()*+,;=:]|%[0-9A-F]{2})*@)?
             (?P<host>([a-z0-9-._~!$&\'()*+,;=]|%[0-9A-F]{2})*)
             (:(?P<port>\d*))?
             (/(?P<path>([^?\#"<>\s]|%[0-9A-F]{2})*/?))?
             (\?(?P<query>([a-z0-9-._~!$&'()*+,;=:/?@]|%[0-9A-F]{2})*))?
             (\#(?P<fragment>([a-z0-9-._~!$&'()*+,;=:/?@]|%[0-9A-F]{2})*))?"""
 
+    URL_DF_REGEX = r"""
```
ianhelle, line 150: Weird - we have a bunch of very similar code (in the defanging PR). Nice!

ianhelle: Or is some of this merged from that PR?

ianhelle (approved):
Nice work! 
A few comments.
Maybe we merge this and then the defang PR, since that has a few extra bits.

docs/source/data_acquisition/SentinelTI.rst
petebryan: Updated this here (and in all other Sentinel docs)

msticpy/context/azure/sentinel_ti.py
petebryan: We can't use all of these as a lot aren't supported by the API. But good idea on aligning. Updated to map across the two

msticpy/context/azure/sentinel_ti.py
petebryan: So you can use the 'keyword' part of the query here if you want to do that. I will make it clearer in the docs. Honestly I don't think this feature is really that useful. I added it as there is an API for it but given the constrained query langauge I don't know why you would use is over runiing KQL against the TIIndicators table.

msticpy/transform/iocextract.py
petebryan: Yeah I merged some stuff for testing the notebook with them, I will merge the rest.

ianhelle (dismissed):

ianhelle (approved):

PR #546, opened 2022-10-28
Add 3.11 to test matrix

tonybaloney:


ianhelle:
Still seems to be failing on statsmodels installs. :-(
Although this is progress - I think it previously died at scipy.
We should track this issue https://github.com/statsmodels/statsmodels/issues/8287

ianhelle (approved):

Issue #263, opened 2022-01-06
Add support for contextual queries to Service Now API


FlorianBracq:
To add context to data extracted from DataProviders, in addition to the existing TIProvider and GeoIpLookup, it would be nice to add contextual data from internal CMDB such as Service Now.

Most HTTP API can be queried from HttpProvider but should not be limited to IOCs supported by TIProvider.

Proposed solution:
* Define a new HttpProvider class under msticpy\common
* Create a folder _context_ under msticpy\data
* Add support for Service Now queries under msticpy\data\context
* Service Now context queries should also be registered as Pivot functions (via pivot_register)
* Service Now queries to be implemented:
    * Query host by IPv4
    * Query host by IPv6
    * Query host by host
    * Query user by email address
    * Query user by account name

FlorianBracq:
Built additional classes to clarify dependencies:
* ContextProvider -> TIProvider
* HttpContextProvider (inheriting from ContextProvider + HttpProvider) -> HttpTIProvider (inhereting from HttpContextProvider + TIProvider)

Created ContextLookup (abstracted from current TILookup) and started to implement a new version TILookup (inheriting from ContextLookup)

FlorianBracq:
Dependant on #282.

Sample implementation available here > https://github.com/FlorianBracq/msticpy/tree/ServiceNow

FlorianBracq:
Hi @ianhelle,

I agree that it's the right time to get on it again, but I'm not sure to follow your idea?

The way I see it, [http_lookup.py](https://github.com/microsoft/msticpy/blob/v2.0.0/msticpy/context/tiproviders/http_lookup.py) being generic/extensible would be more aligned with a parent class (and the class HttpTIProvider should inherit from it, rather than HttpLookupProvider inheriting from HttpTIProvider.

I would see the mid term solution as:
* msticpy/context/
  * base_provider
  * base_http
  * TI_Lookup
  * context_lookup
  * contextproviders/
    * contextprovider (inheriting from base_provider)
    * http_context_provider (inheriting from base_http and contextprovider)
  * tiproviders/
    * tiprovider (inheriting from base_provider)
    * http_context_provider (inheriting from base_http and tiprovider)

and the long term solution would be for base_http to move to common in order for data_providers too to rely on the same code base?

What do you think?

PR #556, opened 2022-11-02
Add support for context provider Service Now

FlorianBracq:
Requires https://github.com/microsoft/msticpy/pull/511

Fixes https://github.com/microsoft/msticpy/issues/263


ianhelle (approved):

ianhelle:
I can add a mechanism to add these as pivots. Looks awesome. 
I have no way of testing this but, it's very cool that we've finally got it in (almost finally)

PR #580, opened 2022-11-22
Updated M365D and MDE data connectors with correct scopes when using delegated auth.

petebryan:
Resolves issue with delegated auth to M365D APIs. Change assigns correct scope to requested token so that the API can be used.


conda/conda-reqs.txt
```diff
@@ -17,7 +17,7 @@ dnspython>=2.0.0, <3.0.0
 folium>=0.9.0
 geoip2>=2.9.0
 html5lib
-httpx>=0.23.1
+httpx==0.23.0
```
ianhelle, line 20: Is there a problem with 0.23.1? I know it needs respx bumped as well

msticpy/data/drivers/odata_driver.py
```diff
@@ -184,12 +184,8 @@ def connect(
                 authority=authority,
                 username=cs_dict["username"],
                 scopes=self.scopes,
-                auth_type=kwargs["auth_type"]
-                if "auth_type" in kwargs
-                else "interactive",
-                location=cs_dict["location"]
-                if "location" in cs_dict
-                else "token_cache.bin",
+                auth_type=kwargs.get("auth_type", "interactive"),
```
ianhelle, line 187: Is interactive a good default vs. devicecode? I thought this didn't work in AML or anywhere with a remote kernel

requirements-all.txt
```diff
@@ -19,7 +19,7 @@ deprecated>=1.2.4
 dnspython>=2.0.0, <3.0.0
 folium>=0.9.0
 geoip2>=2.9.0
-httpx>=0.23.1
+httpx==0.23.0
```
ianhelle, line 22: as above

requirements.txt
```diff
@@ -10,7 +10,7 @@ deprecated>=1.2.4
 dnspython>=2.0.0, <3.0.0
 folium>=0.9.0
 geoip2>=2.9.0
-httpx>=0.23.1
+httpx==0.23.0
```
ianhelle, line 13: ditto

ianhelle (approved):
The code (as much as I understand it) looks good - had a few questions.

msticpy/data/drivers/odata_driver.py
petebryan: Thats a good point - i will swap for devicecode

PR #592, opened 2022-12-20
Update sentinel_analytics.py list_alert_rules API version

pensivepaddle:
Suggest moving to api_version 2022-11-01-preview (counting on no breaking changes?) for listing alert rules (list_alert_rules) as the current one does not return NRT rules.

Ref https://learn.microsoft.com/en-us/rest/api/securityinsights/preview/alert-rules/list?tabs=HTTP


msticpy/context/azure/sentinel_analytics.py
```diff
@@ -75,7 +75,7 @@ def list_alert_rules(self) -> pd.DataFrame:
 
         """
         return self._list_items(  # type: ignore
-            item_type="alert_rules", api_version="2021-10-01"
+            item_type="alert_rules", api_version="2022-11-01-preview"
```
petebryan, line 78: Looks like this API is out of preview now - should be "2022-11-01"
https://learn.microsoft.com/en-us/rest/api/securityinsights/stable/alert-rules/list?tabs=HTTP

petebryan (changes requested):

msticpy/context/azure/sentinel_analytics.py
ianhelle: I've updated

ianhelle (approved):

PR #596, opened 2023-01-03
Added polling detection module 

danielyates2:
Following discussion in https://github.com/microsoft/msticpy/discussions/515 I've added a polling detection module with an implementation of the method developed in [1]. The implementation is the  `PeriodogramPollingDetector`class in msticpy.analysis.polling_detection.

### References
[1] Heard, N. A. and Rubin-Delanchy, P. T. G. and Lawson, D. J. (2014) Filtering automated polling traffic in computer network flow data. In proceedings of IEEE Joint Intelligence and Security Informatics Conference 2014

PR #601, opened 2023-01-08
Added documentation for the polling detection module

danielyates2:
I've added a jupyter notebook documenting the polling detection module which is stored in the docs/notebooks directory.


ianhelle:
Hey @danielyates2
Looks like a great submission. A couple of questions about this PR.
1. Does it supercede pr #596 ? 
2. Also there seems to be a junit.xml file in the commit. Can you remove that?
If you can confirm 1 we can close the other PR and just review this one.

danielyates2:
> Hey @danielyates2 Looks like a great submission. A couple of questions about this PR.
> 
>     1. Does it supercede pr [Added polling detection module  #596](https://github.com/microsoft/msticpy/pull/596) ?
> 
>     2. Also there seems to be a junit.xml file in the commit. Can you remove that?
>        If you can confirm 1 we can close the other PR and just review this one.

Hey @ianhelle 

Thanks!

1. Yes, it does supercede pr [Added polling detection module  #596](https://github.com/microsoft/msticpy/pull/596), sorry for the duplication
2. I've removed the junit xml file and pushed so hopefully everything should be sorted

If there is anything else you need, please let me know

Cheers
Dan

msticpy/analysis/polling_detection.py
```diff
@@ -0,0 +1,170 @@
...
+                    )
+                )
+
+            p_value = sum(compose)
+
+        p_value = min(p_value, 1)
+
+        return test_statistic, p_value
+
+    def detect_polling(
```
ianhelle, line 104: Wondering if it would be good to have a simplified interface that just accepts a DF and does the date->int conversion, grouping, etc.

```python

def detect_polling(
    self,
   data: pd.DataFrame,
   time_column: str,
   groupby: Union[str, List[str]] = None,
   start: Optional[datetime] = None,
   end: Optional[datetime] = None,
   sample_interval: int = 1
):
    if start and end:
        data = data[(data[time_column] >= start) && ((data[time_column] <= end)]
   data["timestamp_int"] = data[time_column].apply(datetime[time_column]).astype("int64")
   p_start = data["timestamp_int"].min()
   p_end = data["timestamp_int"].max()

   if groupby:
       return data.groupby(groupby).apply(
           lambda x: self.detect_polling(x["timestamp_int"], p_start, p_end, sample_interval)
       )
   return self.detect_polling(data["timestamp_int"], p_start, p_end, sample_interval)
```
           
I think this kind of interface might be more easily used by users with data in DFs.
(entirely untested, off the top of my head code)
Perhaps rename the current detect_polling -> detect_polling_arr or something like that?

I guess we need to return the edge/grouping value in the grouped case.

WDYT?

msticpy/analysis/polling_detection.py
```diff
@@ -0,0 +1,170 @@
...
+        counting_process = Counter(timestamps)
+
+        dn_ = np.array([counting_process[t] for t in time_steps])
+        dn_star = dn_ - len(timestamps) / len(time_steps)
+
+        freq, pxx = signal.periodogram(dn_star)
+
+        max_pxx_freq = freq[np.argmax(pxx)]
+
+        print(
```
ianhelle, line 157: Not sure if printing multiple lines per detection pass might be very noisy (esp if you have lots of edges).
You could return this as part of the result - maybe a tuple or dict?
If called from the DF function I sketched out above you could return the whole thing as a DF with cols:
- concated group ID
- p_value
- start
- end
- dominant interval
- frequency

ianhelle (changes requested):
This is very cool but I think we could make it more accessible for people with a simpler interface that does a lot of the transforms for them and they can just supply a DF.
We might also (perhaps v2) be able to have a simple visualization (I really like the 24hr clocks)
or something that hilighted the edges/groups with low p_values.

Awesome stuff!

ianhelle (approved):

ianhelle (changes requested):
I've got a couple of requests in the comments that would make it more accessible/usable by users.

msticpy/analysis/polling_detection.py
danielyates2: Thanks Ian

No problem, I'll change it to accept a DF and then the results will be added as columns to this. The model results would then just be replicated for each edge

#### Non grouped DF
| src_ip | dst_ip | arrival_time | p_value | freq | dominant_interval |
| -------- | -------- | --------------- | ----------- | ----- | ------------------------ |
| 1.1.1.1 | 2.2.2.2 | 2022-01-01 00:00:01 | 0.01 | 0.1 | 10 |
| 1.1.1.1 | 2.2.2.2 | 2022-01-01 00:00:11 | 0.01 | 0.1 | 10 |
| 1.1.1.1 | 2.2.2.2 | 2022-01-01 00:00:21 | 0.01 | 0.1 | 10 |

#### Grouped DF
 | src_ip | dst_ip | arrival_time | p_value | freq | dominant_interval |
| -------- | -------- | --------------- | ----------- | ----- | ------------------------ |
| 1.1.1.1 | 2.2.2.2 | 2022-01-01 00:00:01 | 0.01 | 0.1 | 10 |
| 1.1.1.1 | 2.2.2.2 | 2022-01-01 00:00:11 | 0.01 | 0.1 | 10 |
| 1.1.1.1 | 2.2.2.2 | 2022-01-01 00:00:21 | 0.01 | 0.1 | 10 |
| 3.3.3.3 | 4.4.4.4 | 2022-01-01 00:00:01 | 0.5 | 0.05 | 20 |
| 3.3.3.3 | 4.4.4.4 | 2022-01-01 00:00:51 | 0.5 | 0.05 | 20 |
| 3.3.3.3 | 4.4.4.4 | 2022-01-01 00:01:01 | 0.5 | 0.05 | 20 |

Hopefully that looks like what you were envisioning?

Cheers
Dan

msticpy/analysis/polling_detection.py
danielyates2: Yes, that makes a lot of sense, I don't think I thought this one through!

I'll be implementing the DF idea above so the `detect_polling_arr` function will return a tuple of **p_value, dominant_frequency, dominant_interval**. 
The columns for these 3 values returned will then be created in the DF

msticpy/analysis/polling_detection.py
ianhelle: Yeah - in fact I was thinking that you could make it more general by using a pandas grouping expression.
If I supply a single column (let's say account_name) it could calculate periodicity of logons for each user.
If I supply a list of columns (2 or more) it would essentially build edges on that column set - e.g. I might have account, host, dest_ip - and it would give periodicity of the tuple of specific account + logged on host + dest IP.
Returning the values columns of the grouping in the DF as you show above. 
I'm sure you get what I am talking about - just wanted to make sure that you weren't restricting it to 2 columns.

msticpy/analysis/polling_detection.py
ianhelle: Sounds good!

msticpy/analysis/polling_detection.py
danielyates2: Yeah we're on the same page :)
The src --> dst pair above was just an example. I think I was confusing matters by getting hung up on edges.

Like in your example above there will be a groupby parameter that the analyst can use. If groupby is `None` it's assumed that all of the timestamps in the dataframe are for a single group/edge/entity

danielyates2: Hey @ianhelle 
I've made the changes and resubmitted my pull request
I decided to remove the start and end parameters as I felt it was adding too much complexity, such as
* How to handle the different possible datetime representations
* If groupby is passed then start and ends could either be single values or a list of values for each group
So by default it will use the min and max timestamp for each edge.

What do you think about this?

ianhelle (approved):

danielyates2:
> @danielyates2 please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> 
> Contributor License Agreement

PR #606, opened 2023-01-11
Fix typo

FlorianBracq:
custom parameter handlers is a dictionary.


petebryan (approved):

petebryan:
Good catch, thanks @FlorianBracq

PR #609, opened 2023-01-19
Adding Pulsedive as Threat Intel provider

fr0gger:
This is a TI module for requesting Pulsedive. The PR include the notebook to demonstrate the use case.

https://pulsedive.com/api/


docs/notebooks/PulsediveLookup.ipynb
```diff
@@ -0,0 +1,962 @@
+{
```
ianhelle, line 1: Any chance that you can create an RST doc from the notebook and add to docs/source/data_acquisition
You will need to tidy it up a bit, but you can generate an RST using nbconvert:
```
jupyter nbcovert --to rst your_notebook.ipynb

```
Name it something like PulseDive.rst
Add a line to link it from docs/source/DataEnrichment.rst
```rst
.. toctree::
   :maxdepth: 2

   data_acquisition/TIProviders
   data_acquisition/GeoIPLookups
   data_acquisition/AzureData
   data_acquisition/Sentinel
   data_acquisition/IPWhois
   data_acquisition/PulseDive
```

To generate the docs:
```
cd docs
make html nogen
# open docs/build/html/index
```
You should be able to see your doc in the ToC under the Enriching section
You should also be able to find the module in the API section.
(although for this to be generated you'll need to do `make html`  - omitting the "nogen")

ianhelle (changes requested):

ianhelle:
@fr0gger 
Thomas, I've hacked around your code a little - mainly to add Pulsedive as a regular MSTICPy TI provider.
I've simplified the interface methods a bit since, in most cases, you don't need to supply a pd_type (e.g. if you are doing a scan or a query, the type is always "scan" or "q") - I've left lookup_ioc with this parm, which just invokes the underlying operation-specific method.
I updated the notebook to match these changes and added a little about the standard TI provider use.

ianhelle (approved):

fr0gger:
Looks good to me, thanks for updating it @ianhelle

PR #622, opened 2023-02-04
local_data_driver: add file_extensions option and json files processing

juju4:
This supports velociraptor offline collection files reading.

PR #623, opened 2023-02-04
add localdata queries file and notebook examples to process velociraptor offline collection

juju4:
* Requires json part for localdata provider - https://github.com/microsoft/msticpy/pull/622
* usual output of velociraptor offline collection can be json or csv. ideally, queries file can include both but it does not seem possible currently


ianhelle (approved):
Looks nice.
I wonder if we want to add a DataEnvironment to match OSQueryLogs. 
We could create an alias for LocalData and call it VelociraptorLogs - this would just let people do
```python
vr_prov = QueryProvider("Velociraptor", data_paths=[...)
```
Then we could tie the VR queries specifically to this provider.
If you think this is a good idea, I can probably implement this.

juju4:
It is a good idea and would be even better if it can be used on top of other queryprovider like sentinel, elastic and so on.
PR about is local logs but if can reuse easily for logs from sentinel & co, the better.

From a quick review, I'm not sure the fail tests are related to PR. comments?

ianhelle:
Superceded by #668

PR #630, opened 2023-02-16
Fix error when latest version 3.0.3 of bokeh is installed

ctoma73:
When "azure" feature is installed the following

import msticpy as mp
mp.init_notebook()

errors with:
ImportError: cannot import name 'Figure' from 'bokeh.plotting' (/usr/lib/python3.9/site-packages/bokeh/plotting/__init__.py)

The fix tries to support both versions of bokeh (<3.0.0) as well as latest version of bokeh using aliasing.


ianhelle (approved):

msticpy/vis/entity_graph_tools.py
```diff
@@ -15,6 +15,8 @@
 from bokeh.models import Circle, HoverTool, Label, LayoutDOM
 from bokeh.plotting import figure, from_networkx
 
+from msticpy.vis.figure_dimension import figure_dimension
```
ianhelle, line 18: We use relative imports for in-package imports.
```python
from .figure_dimension import figure_dimension
```

msticpy/vis/figure_dimension.py
```diff
@@ -0,0 +1,20 @@
+# -------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for
+# license information.
+# --------------------------------------------------------------------------
+
+from bokeh.plotting import figure
+
+class figure_dimension:
```
ianhelle, line 9: Actually, on second pass, does this need to be class?
Should just be a function, right?

msticpy/vis/network_plot.py
```diff
@@ -24,6 +25,9 @@
 from bokeh.plotting import figure, from_networkx, show
 from typing_extensions import Literal
 
+from msticpy.transform.network import df_to_networkx
```
ianhelle, line 28: use relative imports

msticpy/vis/timeline.py
```diff
@@ -15,6 +15,8 @@
 from bokeh.models.annotations import LegendItem
 from bokeh.plotting import figure, reset_output
 
+from msticpy.vis.figure_dimension import figure_dimension
```
ianhelle, line 18: relative imports

msticpy/vis/timeline_common.py
```diff
@@ -22,7 +22,16 @@
 from bokeh.palettes import Palette, viridis
 
 # pylint: enable=no-name-in-module
-from bokeh.plotting import Figure, figure
+from bokeh.plotting import figure
+
+from msticpy.vis.figure_dimension import figure_dimension
```
ianhelle, line 27: ditto

msticpy/vis/timeline_duration.py
```diff
@@ -17,6 +17,8 @@
 from bokeh.plotting import figure, reset_output
 from bokeh.transform import dodge
 
+from msticpy.vis.figure_dimension import figure_dimension
```
ianhelle, line 20: ditto

msticpy/vis/timeline_values.py
```diff
@@ -14,6 +14,8 @@
 from bokeh.models import ColumnDataSource, HoverTool, LayoutDOM, Legend
 from bokeh.plotting import figure, reset_output
 
+from msticpy.vis.figure_dimension import figure_dimension
```
ianhelle, line 17: ditto

msticpy/vis/timeseries.py
```diff
@@ -16,6 +16,8 @@
 from bokeh.plotting import figure, reset_output
 from deprecated.sphinx import deprecated
 
+from msticpy.vis.figure_dimension import figure_dimension
```
ianhelle, line 19: ditto

ianhelle (changes requested):
I ran through it a bit too quickly.
Minor things though - I think the class should be function and use of relative imports

ianhelle (approved):

PR #632, opened 2023-02-23
Ianhelle/mp extensibility 2023 02 09

ianhelle:
This is mainly adding functionality to be able to "side-load" data providers, TI and context providers.
Implemented this in mp_plugins.py
Added unit test with some sample providers
Have made some small changes to QueryProviders, QuerySource and TI/Context provider infrastructure for this.

I've documented this in a new RTD page and took a bit of time to re-org and write
some supplemental documentation.
We now have a Developer section in the docs - mainly with content taken from the repo Wiki
We also have an Extending MSTICPy section
- this has a page documenting the plugin model
- I've added a new document describing how to build a TI or Context provider
- I've moved the Writing a Data Provider to this section
- I've also move the Creating Queries and Creating Pivot functions sections to here and
  left links in the original docs.

## Writing a TI provider or Context provider

```python

    class TIProviderHttpTest(HttpTIProvider):
        """Custom IT provider TI HTTP."""

        PROVIDER_NAME = "MyTIProvider"
        _BASE_URL = "https://api.service.com"
        _QUERIES = _QUERIES = {
            "ipv4": APILookupParams(path="/api/v1/indicators/IPv4/{observable}/general"),
            "ipv6": APILookupParams(path="/api/v1/indicators/IPv6/{observable}/general"),
```

## Using plugins with MSTICPy

Load on demand
```python

    import msticpy as mp

    mp.load_plugins(plugin_paths="/my_modules")

    # or multiple paths
    mp.load_plugins(
        plugin_paths=["./my_modules", "./my_other_modules"]
    )
```

Or specify in `msticpyconfig.yaml`

```yaml
        ...
        Custom:
            - "testdata"
    PluginFolders:
        - tests/testdata/plugins
    Azure:
        ...
```


docs/source/extending/PivotFunctions.rst
```diff
@@ -0,0 +1,270 @@
...
++-------------------------+-------------------------------+------------+------------+
+| func_static_params      | dict of static name/value     | No         | -          |
+|                         | params always sent to the     |            |            |
+|                         | function                      |            |            |
++-------------------------+-------------------------------+------------+------------+
+| func_input_value_arg    | Name of the param that the    | If not     | -          |
+|                         | wrapped function uses for its | DF input   |            |
+|                         | input value                   |            |            |
++-------------------------+-------------------------------+------------+------------+
+| can_iterate             | True if the function supports | No         | Yes        |
```
rcobb-scwx, line 68: Do pivot functions with `can_iterate=True` behave like this?

```python
# .get_recent_alerts would be the custom pivot function
for recent_alert in IpAddress.get_recent_alerts("157.53.1.1"):
    ...
```

Just trying to understand how they would be used.

docs/source/extending/WritingTIAndContextProviders.rst
```diff
@@ -0,0 +1,460 @@
...
+        ),
+        ...
+
+**data**
+This is currently not supported but we will implement if and
+when required. This is a dictionary that will be supplied as request data. supports
+parameter substitution for values.
+
+**auth_type**
+Currently only "HTTPBasic" is supported.
```
rcobb-scwx, line 253: This isn't limited to HTTP basic auth in the technical sense (ie, b64 encoded username:password). It can handle any URL parameter or HTTP header based authentication if I am reading the docs correctly.

docs/source/extending/WritingTIAndContextProviders.rst
```diff
@@ -0,0 +1,460 @@
...
+The implementation in the example at the start of this document
+(and below) shows a simple process, but it can be as complex as needed.
+
+.. note:: we would recommend creating child methods to handle
+    different response types if you need to do complex
+    processing.
+
+.. code:: python3
+
+    def parse_results(self, response: Dict) -> Tuple[bool, ResultSeverity, Any]:
```
rcobb-scwx, line 290: Is `Tuple[bool, ResultSeverity, Any]` the required return type or just what you've chosen for this example? If the former, then creating/re-using a new type would be easier than copy/pasting that annotation for each `parse_results` method.

rcobb-scwx (commented):
I think this PR would bring much-needed extensibility to `msticpy`. Thanks for putting this together. Here are a few high-level thoughts (some of this may have been previously discussed, so my apologies if it is redundant or this is the wrong place to bring them up):

In theory, creating/registering custom pivot functions should be easier than the plugins because they are just configuration options in YAML, rather than bespoke code. But looking at the number of configuration options and how deeply related they are to Python objects and internals, I am not sure if the bar-to-entry is much lower for the configuration approach.

In either case - extensions as code or extensions as configuration - the extension developer is ultimately responsible for making sure whatever custom code/functions are importable at runtime. These changes don't provide extension developers with best practices or guidance on how to ship the extensions to their users. For example, you can configure custom pivot functions that reference some function in a proprietary security vendor SDK, but having the configuration YAML(s) is not sufficient to actually use the extension; there is an extra package management step that is still required here.

On a similar note with how plugins are loaded: I still think namespace packages are the way forward for loading custom extensions. Yes, defining an extension as a proper python package requires a lot of boilerplate -- but you get the benefit of handling additional dependencies and auto-discovery of installed extensions in the namespace. I think a cookiecutter skeleton project showing how to write a msticpy extension would fix the boilerplate issue and encourage better development practices.

Just my 2c. Hope some of my comments were helpful!

docs/source/extending/PivotFunctions.rst
ianhelle: It really means that the function is not too expensive to run. When you supply an iterable of DF input to pivot functions, if the wrapped (underlying) function only accepts single items, the pivot stuff will call the wrapped function multiple times and concat the results. This is fine for most things but some things like notebooklets are big and expensive and I wanted to avoid iterating over those by accident.

docs/source/extending/WritingTIAndContextProviders.rst
ianhelle: Ok. If you can suggest better wording, that would be great. Just that the auth param we supply to httpx is "HTTPBasic" vs. a bunch of other mechanism like client cert. Agree that the text is prob misleading.

docs/source/extending/WritingTIAndContextProviders.rst
ianhelle: Great idea! I normally now do that as a rule (returning raw tuples is nasty so I usually use named tuples), but this a hangover of prehistoric code from the early days of msticpy.
I'll implement that.

docs/source/extending/Queries.rst
```diff
@@ -0,0 +1,753 @@
...
+                type: datetime
+        query_macros:
+            query_condition:
+                description: Query-specific where clause
+                value: '| where (VMIP_s in ({ip_address_list})
+                or SrcIP_s in ({ip_address_list})
+                or DestIP_s in ({ip_address_list})
+                )'
+
+This allows you define a series of related queries that have the
```
FlorianBracq, line 369: s/you define/you to define/

docs/source/extending/Queries.rst
```diff
@@ -0,0 +1,753 @@
...
+Once imported the queries in the files appear in the Query Provider's
+Query Store alongside the others and can be called in the same manner as
+pre-defined queries.
+
+If you have created a large number of query definition files and you
+want to have the automatically imported into a Query Provider's query
+store at initialization you can specify a directory containing these
+queries in the msticpyconfig.yaml file under QueryDefinitions: Custom:
+
+For example, if I have a folders with queries in two folders
```
FlorianBracq, line 498: I'm not sure what was meant here?

docs/source/extending/WritingTIAndContextProviders.rst
```diff
@@ -0,0 +1,460 @@
...
+
+Context providers follow the same model as TI Providers.
+
+The key differences are as follows.
+
+parse_results method
+~~~~~~~~~~~~~~~~~~~~
+
+This method is used only to extract details - unlike the TI providers, there
+is no severity scoring. It should a tuple of (*Success*, *Details*) - where
```
FlorianBracq, line 440: s/should a/should be a/

msticpy/init/mp_plugins.py
```diff
@@ -0,0 +1,115 @@
...
+_PLUGIN_TYPES = {
+    DriverBase: PluginReg(drivers, "DATA_ENVIRONMENTS"),
+    TIProvider: PluginReg(TILookup, "PROVIDER_NAME"),
+    ContextProvider: PluginReg(ContextLookup, "PROVIDER_NAME"),
+}
+
+
+def read_plugins(plugin_paths: Union[str, Iterable[str]]):
+    """Load plugins from folders specified in msticpyconfig.yaml."""
+    plugin_config = [plugin_paths] if isinstance(plugin_paths, str) else plugin_paths
```
FlorianBracq, line 72: In case the type changes,  I would implement this the other way:
```
plugin_paths if isinstance(plugin_paths, list) else [plugin_paths]
```

FlorianBracq (commented):
Great work on the documentation front! That definitely helps understanding how the code is organized and how each parts works with the other.
I did not review too much on the implementation as the "namespace" implementation looks helpful and might be more straightforward to implement.

docs/source/extending/PluginFramework.rst
```diff
@@ -0,0 +1,136 @@
...
+name of your provider class. You can customize this by adding
+a ``DATA_PROVIDERS`` attribute to your class. This should
+be a list of strings.
+
+.. code:: python3
+
+    class CustomDataProvB(CustomDataProvA):
+        """Custom provider."""
+
+        DATA_ENVIRONMENTS = ["SQLTestProvider", "SQLProdProvider"]
```
rcobb-scwx, line 72: Should this be DATA_PROVIDERS or DATA_ENVIRONMENTS? See line 64 above

docs/source/extending/PluginFramework.rst
```diff
@@ -0,0 +1,138 @@
...
+enrichment sources.
+
+These currently include:
+
+- Data Providers
+- Threat Intelligence Providers
+- Context Providers
+
+If you create one of these and think that it would be useful
+to other users, please consider contributing it to MSTICPy.
```
petebryan, line 15: Maybe we could link to contribution guidance here or even make the language a bit stronger such as "If you are creating an extension that others can use consider contributing it to MSTICPy, however if you are creating something specific to you or your organization then you can create local provider modules.

docs/source/extending/Queries.rst
```diff
@@ -0,0 +1,755 @@
...
+            end:
+                description: Query end time
+                type: datetime
+        query_macros:
+            query_condition:
+                description: Query-specific where clause
+                value: '| where (VMIP_s in ({ip_address_list})
+                or SrcIP_s in ({ip_address_list})
+                or DestIP_s in ({ip_address_list})
+                )'
```
petebryan, line 367: i'm not sure this example makes it clear enough why someone would use the macro here.

docs/source/extending/WritingTIAndContextProviders.rst
```diff
@@ -0,0 +1,460 @@
...
+document is applicable to both so you should read the entire
+thing. A short section on the differences for Context providers
+follows this first section.
+
+
+Threat Intelligence Providers
+-----------------------------
+
+For most TI services you can use
+:py:class:`HttpTIProvider <msticpy.context.tiproviders.ti_http_provider.HttpTIProvider>`.
```
petebryan, line 25: I think we should make clear this is used where your TIP offers an API you can query over the web as not all TIPs will have this.

petebryan (approved):

PR #633, opened 2023-02-24
Modified the upload_df method to split the data into batches of 10,00…

jllangley:
-  Modified the upload_df method to split the data into batches of 10,000 rows and upload each batch to Log Analytics separately. This should help to improve the upload speed and avoid timeouts for large data sets.
-  Added a new argument `batch_size` to the method signature to control the size of batches when uploading the dataframe to Log Analytics.


jllangley:
> 

ianhelle (approved):

ianhelle:
Thanks @jllangley for this!

PR #634, opened 2023-02-27
Improved Splunk driver with the JSON reader class and a new progress bar

cyberfiredragon:
Here is a improved Splunk driver with the following changes:
- It uses the new JSON results reader (recommended splunklib 1.7.2, but available since 1.6.12)
- It uses new progress bar: alive_progress (https://pypi.org/project/alive-progress/)
- It supports access via Web Proxy (it is very important for many enterprises), parameters are passed via the "connect" string


cyberfiredragon:
I will provide an updated version shortly

PR #640, opened 2023-03-14
Adding logging and updating settings access

ianhelle:
Added Python logging to MSTICPy - currently only a few modules have this implemented but it's now there for use.
We should instrument places like the authentication modules.
I've also changed the settings access so that everything goes through get_config, set_config. First steps to making testing with different configs easier.


msticpy/init/azure_synapse_tools.py
```diff
@@ -468,33 +487,65 @@
...
+    logger.info(
+        "Setting authentication env var %s=%s",
+        AzureCredEnvNames.AZURE_CLIENT_ID,
+        client_id,
     )
-    os.environ[AzureCredEnvNames.AZURE_CLIENT_SECRET] = mp_spark.get_kv_secret(
-        SynapseName.sp_client_sec_name
+    logger.info(
+        "Setting authentication env var %s=%s",
+        AzureCredEnvNames.AZURE_CLIENT_SECRET,
```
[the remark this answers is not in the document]
ianhelle: The secret value is blanked - only asterisks are logged.

msticpy/data/drivers/kql_driver.py
```diff
@@ -420,15 +420,13 @@ def _set_kql_option(option, value):
...
-        print(kql_config)
         current_opts = {
             opt.split("=")[0].strip(): opt.split("=")[1]
             for opt in kql_config.split(";")
         }
-        print(current_opts)
         current_opts[option] = value
         kql_config = ";".join(f"{opt}={val}" for opt, val in current_opts.items())
-        print(kql_config)
+        # print(kql_config) - replace with logger
```
petebryan, line 429: Was this a placeholder?

petebryan (approved):

msticpy/data/drivers/kql_driver.py
ianhelle: It was - there were a few but I thought I'd leave one there as a reminder of places to add logging.

PR #645, opened 2023-03-17
Added multithreading support for additional connections (+fixes)

d3vzer0:
# Support for running a query across multiple connections (with optional async operation)

It is common for data services to be spread across multiple tenants or workloads. E.g., multiple Sentinel workspaces,
Microsoft Defender subscriptions or Splunk instances. You can use the ``QueryProvider`` to run a query across multiple connections and return the results in a single DataFrame.

To create a multi-instance provider, create an instance of a QueryProvider for your data source and execute the `connect()` method to connect to the first instance of your data service. 

Then use the `add_connection()` method. This takes the same parameters as the `connect()` method (the parameters for this method vary by data provider).

`add_connection()` also supports an ``alias`` parameter to allow you to refer to the connection by a friendly name. 

```python3

    qry_prov = QueryProvider("MSSentinel")
    qry_prov.connect(workspace="Workspace1")
    qry_prov.add_connection(workspace="Workspace2, alias="Workspace2")
    qry_prov.list_connections()

```
When you now run a query for this provider, the query will be run on all of the connections and the results will be returned as a single dataframe.

```python3

    test_query = '''
        SecurityAlert
        | take 5
        '''

    query_test = qry_prov.exec_query(query=test_query)
    query_test.head()
```

Some of the MSTICPy drivers support asynchronous execution of queries against multiple instances, so that the time taken to run the query is much reduced compared to running the queries sequentially. Drivers that support asynchronous queries will use this automatically. The initial set of multi-threaded drivers are:
- MSSentinel_New (the new version of the MSSentinel driver)
- Kusto_New (the new version of the Kusto/Azure Data Explorer driver)

By default, the queries will use at most 4 concurrent threads. You can override this by initializing the QueryProvider with the
`max_threads` parameter to set it to the number of threads you want.

```python3

    qry_prov = QueryProvider("MSSentinel", max_threads=10)
```

# Multi-threaded support for split/shared queries

MSTICPy has supported splitting large queries by time-slice for a while. This is documented here [Splitting a Query into time chunks](https://msticpy.readthedocs.io/en/latest/data_acquisition/DataProviders.html#splitting-query-execution-into-chunks). With this release, we've added asynchronous support for this (if the driver supports threaded/async operation) so that multiple chunks of the query will run in parallel.

```python3

    qry_prov.SecurityAlert.list_alerts(start=start, end=end, split_by="1d")

```
Use the parameter `split_query_by` or `split_by` to specify a time range (the time unit uses the same syntax as pandas time intervals - e.g. "1D", "4h", etc. - the the pandas documentation for more details on this).

In this release sharding is also supported for ad hoc queries as long as you add "start" and "end" parameters to the query (this is still experimental, so let us know if you have issues with this).


msticpy/data/core/data_providers.py
```diff
@@ -193,12 +196,13 @@ def add_connection(
 
         """
         # create a new instance of the driver class
-        new_driver = self.driver_class(**(self._driver_kwargs))
-        # connect
+        new_driver = QueryProvider(self.environment,
```
d3vzer0, line 199: This was changed since the previous method to initialise he driver resulted in the incorrect dataprovider being used. Ie. when initialising `QueryProvider("M365D")` every additional connection used a different API endpoint for the queries (securitycenter.microsoft.com) since `self.data_environment` was set to none (for the additional connections).

msticpy/data/core/data_providers.py
```diff
@@ -449,22 +453,39 @@ def exec_query(self, query: str, **kwargs) -> Union[pd.DataFrame, Any]:
         """
         query_options = kwargs.pop("query_options", {}) or kwargs
         query_source = kwargs.pop("query_source", None)
-        result = self._query_provider.query(
-            query, query_source=query_source, **query_options
-        )
+
         if not self._additional_connections:
+            result = self._query_provider.query(
```
d3vzer0, line 458: Moved the 'single' query here to fix an issue where executing a query required a connect() even though multiple connections were added via the `add_connection` function

msticpy/data/core/data_providers.py
```diff
@@ -449,22 +453,39 @@ def exec_query(self, query: str, **kwargs) -> Union[pd.DataFrame, Any]:
...
-        for con_name, connection in self._additional_connections.items():
-            print(f"{con_name}...")
-            try:
-                results.append(
-                    connection.query(query, query_source=query_source, **query_options)
-                )
-            except MsticpyDataQueryError:
-                print(f"Query {con_name} failed.")
+        if self._query_provider._support_async:
+            with ThreadPoolExecutor(max_workers=self._query_provider.max_threads) as executor:
```
d3vzer0, line 467: Not sure if this is the 'prettiest' way to do this since there are duplicate lines between the threaded variant and the sequential variant

msticpy/data/core/data_providers.py
ianhelle: Makes sense

msticpy/data/core/data_providers.py
ianhelle: I think that makes sense. Definitely.

msticpy/data/core/data_providers.py
ianhelle: I think it's prob OK. I think it would more confusing to try to refactor out the common items.
One change I would make (I know it was my name originally) is change "MsticpyInstance" to "DataProviderInstance"

ianhelle (dismissed):
I think the only change that would be good is changing MsticpyInstace to DataProviderInstance.
Looks great though!!!

ianhelle:
@d3vzer0 - I'm going to make some additions to this to add the same capability for splitting queries by time.
I've made some refactoring changes in PR #656 - I want to put all of the multi-connection functionality in a separate mixin class. So I'll get back to this when this PR is merged and synced with this one.
Awesome stuff though - thanks for the work.

ianhelle:
Hey @d3vzer0,
Inspired by the PR, I implemented this for both multiple connections and for split (by time interval) queries.
Since the QueryProvider class was getting a bit out of hand, I moved all of the connections/threading code into its own mixin class. I haven't full tested this yet but it seems to work for both scenarios.
I've also added this all to the docs so that it's visible to more than just you and me. :-D
Let me know what you think.

msticpy/data/core/query_provider_connections_mixin.py
```diff
@@ -81,18 +101,361 @@ def list_connections(self) -> List[str]:
...
+                self.exec_query,
+                query=query_str,
+                query_source=query_source,
+                time_span={"start": start, "end": end},
+                **query_options,
+            )
+            for (start, end), query_str in split_queries.items()
+        }
+        # Run the queries threaded if supported
+        if self._query_provider.get_driver_property(DriverProps.SUPPORTS_THREADING):
```
rcobb-scwx, line 286: This block seems repetitious with https://github.com/microsoft/msticpy/pull/645/files#diff-8c03afafd2896629d2c9a856c9416072a595ba61621d02f3e5427cf267b692afR156. Might be an easy refactor into a separate function.

msticpy/data/drivers/mdatp_driver.py
```diff
@@ -40,6 +42,7 @@ def __init__(self, connection_str: str = None, instance: str = "Default", **kwar
 
         """
         super().__init__(**kwargs)
+        self._support_async = True
```
rcobb-scwx, line 45: Should we use  `DriverProp.SUPPORTS_ASYNC` instead to be consistent with the threaded approach?

ianhelle: Yep - absolutely. I think I missed this

ianhelle (approved):

PR #660, opened 2023-04-29
Validation of the YAML structure of query files

FlorianBracq:
### Goal of the PR
Add through library jsonschema an automated validation of the query files.

### Details
The structure of a query file is defined in the JSON file; .schemas\queries.json

For now, the validation is fairly lenient, but more checks can be added, such as ensuring that when a key is used, a value is assigned to it.
I've noticed a few existing queries defining a key "metadata" without setting any metadata for instance; should it be considered invalid?

Also, it integrates nicely with VSCode as you can (after installing the VSCode extension "YAML" by RedHat) configure your editor to automatically highlight YAML structure errors, or autocomplete by adding in the workspace's settings.json:
```
    "yaml.schemas": {
        ".schemas/queries.json": [
            "/msticpy/data/queries/**/*.yaml"
        ]
    },
```


ianhelle:
Hey Florian - weird coincidence. I've been playing around with Github CopilotX to generate code. One of the things was UI for creating queries and experiments with using Pydantic models for the query files. I think we could use this to generate the json schema for the checker.

FlorianBracq:
Generating a similar schema for the msticpyconfig file should not that hard if you can provide a 'complete' example of all the options. That's what stopped me here as I was not able to find that list and was afraid it would break something?

Another point I was not sure about was when should this validation be executed?
* from vscode with the user 'manually opting-in' by configuring his/her workspace sounded good
* from the unitary tests sounded reasonable too
* when importing msticpy to confirm all queries are valid sounded possible, but I was not entirely sure it would be a good idea?

Happy to receive your feedback on this!

ianhelle:
Maybe we should have talked about this....to avoid a bit of duplication. I started down the json schema route and created a schema for msticpyconfig but then went to have a look at Pydantic (Ryan is a big fan). On the whole this seems much easier to create and maintain since you define the schema in terms of Python classes.
I have a couple of relevant branches:
- https://github.com/microsoft/msticpy/tree/ianhelle/include-files-in-config-2023-04021 - this contains a pydantic model file for the config file.
- https://github.com/microsoft/msticpy/tree/ianhelle/query-editor-2023-04-21 - this one contains an object model built in dataclasses. I used this to generate the UI code for the editor (using CoPilotX). pydantic has drop-in replacements for dataclass - I was thinking of making this a pydantic model as well and using it to validate the queries.

As far as when to do the validation.
- For queries, it doesn't make sense to do this before importing them (triggered by loading a specific driver/provider type) but it does make sense then. In all likelihood the code would crash when trying to read a bad query file but maybe not. There is some simple validation done in the query reader file.
- For the msticpyconfig - it makes sense to do this when reading it (i.e. when importing msticpy) but I also wanted to make a more verbose checker for helping diagnose user issues (e.g. you could have a syntantically-correct file but no Sentinel workspaces defined, which would obv be a problem if you were trying to connect to Sentinel).

I'm now hesitant about doing the pydantic thing for query files since it duplicates this PR. Even if we went the pydantic route it would be good to generate json schemas for both file types for use in vscode and elsewhere.

.schemas/queries.json
```diff
@@ -0,0 +1,262 @@
...
+                    "type": "integer"
+                },
+                "description": {
+                    "$ref": "#/$defs/description"
+                },
+                "data_environments": {
+                    "type": "array",
+                    "items": {
+                        "type": "string",
+                        "enum": [
```
ianhelle, line 72: Is it possible to have this list open-ended? E.g. for custom providers.
Maybe not that important for our use of the json file.
It's also missing a couple of recent providers - MSSentinel_New and Kusto_New

ianhelle (changes requested):
Looks good overall. 
I have a PR open for a query editor implemented in ipywidgets. This has a pydantic data model for the query structure. Not sure if this might be in conflict with some of the requirements of the json schema.

msticpy/data/queries/localdata/local_data.yaml
```diff
@@ -7,68 +7,58 @@ metadata:
 defaults:
   metadata:
     data_source: 'security_alert'
-  parameters:
```
ianhelle, line 10: Does validation fail if there is an empty "parameters" element?

msticpy/data/queries/m365d/kql_m365_hunting.yaml
```diff
@@ -509,5 +491,4 @@ sources:
         | where DownloadTime - {time_column} between (0min .. 2min)
         | project-away DeviceName1
         {add_query_items}'
-      uri: "https://github.com/microsoft/WindowsDefenderATP-Hunting-Queries/blob/master/Protection%20events/SmartScreen%20URL%20block%20ignored%20by%20user.txt"
```
ianhelle, line 512: We can delete the uri element if it's causing an issue - it was an idea that was never used.

ianhelle (changes requested):
Couple of comments.

.schemas/queries.json
FlorianBracq: The list can be open ended thanks to the "anyOf" as explained here > https://github.com/json-schema-org/json-schema-spec/issues/665
However that effectively means the list of valid providers will only be there for autocompletion when integrated with vscode (or another text editor).

I'll implement the change for this

tests/common/test_pkg_config.py
```diff
@@ -31,10 +49,26 @@ def test_load_default():
...
+        path = Path(path)
+        if not path.is_absolute():
+            path = Path(__file__).resolve().parent.parent.joinpath(path)
+        check.is_true(path.is_dir())
+        for query_file in Path(path).resolve().rglob("*.yaml"):
+            validate_queries_file_structure(
+                query_file,
+                expected=not (
+                    "fail" in query_file.name
+                    or "sentinel_query_import_data" in query_file.absolute
```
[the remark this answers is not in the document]
FlorianBracq: damn, you are right! sorry about that

msticpy/data/queries/localdata/local_data.yaml
FlorianBracq: As parameters is a dictionary (or an "object" in the JSON validation word) and was defined to contain at least 2 required keys (description & type). Since those keys are required, parameters cannot be empty. By removing this requirement, we could have empty parameters, but also parameters without a description or a type.

Happy to change it if you believe that's a better approach.

https://json-schema.org/understanding-json-schema/reference/object.html#id3

msticpy/data/queries/m365d/kql_m365_hunting.yaml
FlorianBracq: The uri field is not the issue, it's just git messing up the details of the change. The uri field has not been modified but is now the new "last line of the file"

ianhelle (approved):

tests/common/test_pkg_config.py
```diff
@@ -31,10 +49,27 @@ def test_load_default():
...
-        path = f"data/{path}"
-        check.is_true(
-            Path(pkg_config.__file__).resolve().parent.parent.joinpath(path).is_dir()
+        path = (
+            Path(pkg_config.__file__).resolve().parent.parent.joinpath(f"data/{path}")
         )
+        check.is_true(path.is_dir())
+        for query_file in path.rglob("*.yaml"):
+            validate_queries_file_structure(query_file)
+    for path in settings["QueryDefinitions"].get("Custom", []):
```
ianhelle, line 58: I think it's maybe risky to be checking all the testdata files - since there may be some query files that are intentionally malformed.
The test is failing because it's getting "testdata" as the custom path from the msticpyconfig-test.yaml and one of the test files for the sentinel query downloaded is a query when the test code expects it not to be.
Seems like we only really want to test prod query files that are in the msticpy/data/queries folder. I think we can remove the second part of this test. Maybe can add something back here or build a special test with controlled msticpyconfig.yaml paths if we ever need to test query files in a custom folder.

tests/common/test_pkg_config.py
```diff
@@ -9,20 +9,38 @@
...
-from ..unit_test_lib import custom_mp_config, get_test_data_path
+from ..unit_test_lib import custom_mp_config, get_test_data_path, get_queries_schema
 
 _TEST_DATA = get_test_data_path()
+_QUERIES_SCHEMA = get_queries_schema()
 
 # pylint: disable=protected-access
 
 
+def validate_queries_file_structure(query_file: Path, expected: bool = True):
```
ianhelle, line 28: I think this really should be a test in the tests/data folder as its own test module..
The pkg_config tests are meant to test for people adding dependencies that are not set in setup.py or requirements (and  ensure that the requirements files all agree).

ianhelle (changes requested):
I missed a couple of items last time.
Looks good otherwise though (better than good :-) )

tests/common/test_pkg_config.py
FlorianBracq: I understood why this test was failing. I did not define any of the "root keys" (ie: metadata, sources and defaults) to be required. Therefore, the Sentinel rules somehow matched the definition since it did not contain any of those keys...

I was also using this test to validate my own "custom queries" defined in my msticpyconfig file, but I do not know if that's a valid case?

tests/common/test_pkg_config.py
FlorianBracq: Happy to add it as its own test module, but what test would like to see? I could add a few good & bad query files and check if they match the expected output?

ianhelle (approved):

tests/common/test_pkg_config.py
ianhelle: Maybe create something in tests/data/queries/test_query_files.py

tests/common/test_pkg_config.py
ianhelle: Hmm that seems valid. however there are or could be invalid query files in the tests\testdata folder.
I guess in your test you could remove the custom queries key from the config or use a custom msticpyconfig.
Maybe easier just to skip the path in your test if it points to the tests folder.

tests/common/test_pkg_config.py
FlorianBracq: New test file created with sample query files!

tests/common/test_pkg_config.py
FlorianBracq: Content from the test folder is now ignored from the package validation

PR #668, opened 2023-05-19
Ianhelle/velociraptor provider 2023 05 19

ianhelle:
# Velociraptor data provider

The ``Velociraptor`` data provider can read Velociraptor log files and provide convenient query functions for each data set in the output logs.

The provider can read files from one or more hosts, stored in in separate folders. The files are read, converted to pandas DataFrames and grouped by table/event. Multiple log files of the same type (when reading in data from multiple hosts) are concatenated into a single DataFrame.

To use the Velociraptor provider, you need to create an `QueryProvider` instance, passing the string "Velociraptor" (or "VelociraptorLogs") as the `data_environment` parameter. You also need to add the `data_paths` parameter to specify specific folders that you want to search for log file (although you can set these paths in msticpyconfig.yaml, if you do this frequently).

You can specify multiple folders to have the logs from different hosts.

```python3
    qry_prov = mp.QueryProvider("VelociraptorLogs", data_paths=["~/my_logs"])
```
Calling the `connect` method triggers the provider to read the locations of the
log files (although the contents are not read until a query function is run).

```python3

    qry_prov.connect()


## Listing Velociraptor tables

```python3
    qry_prov.list_queries()
```
```
    ['velociraptor.Custom_Windows_NetBIOS',
    'velociraptor.Custom_Windows_Patches',
    'velociraptor.Custom_Windows_Sysinternals_PSInfo',
    'velociraptor.Custom_Windows_Sysinternals_PSLoggedOn',
   ....
```

Each query returns the table of data types retrieved from the logs.

```python3

    qry_prov.vc_prov.velociraptor.Windows_Forensics_ProcessInfo()
```

| Name  | PebBaseAddress   |   Pid | ImagePathName | CommandLine  | CurrentDirectory     | Env   |                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             
| :------  | :---------------       | ----: | :-----------           | :---------------- | :-----------------  | :---- |
| LogonUI.exe | 0x95bd3d2000     |   804 | C:\Windows\system32\LogonUI.exe | "LogonUI.exe" /flags:0x2 /state0:0xa3b92855 /state1:0x41c64e6d   | C:\Windows\system32\ | {'ALLUSERSP |
| dwm.exe     | 0x6cf4351000     |   848 | C:\Windows\system32\dwm.exe     | "dwm.exe"                                                        | C:\Windows\system32\ | {'ALLUSERSP |
| svchost.exe | 0x6cd64d000      |   872 | C:\Windows\System32\svchost.exe | C:\Windows\System32\svchost.exe -k termsvcs                      | C:\Windows\system32\ | {'ALLUSERSP |
| svchost.exe | 0x7d18e99000     |   912 | C:\Windows\System32\svchost.exe | C:\Windows\System32\svchost.exe -k LocalServiceNetworkRestricted | C:\Windows\system32\ | {'ALLUSERSP |
| svchost.exe | 0x5c762eb000     |   920 | C:\Windows\system32\svchost.exe | C:\Windows\system32\svchost.exe -k LocalService                  | C:\Windows\system32\ | {'ALLUSERSP |


msticpy/data/drivers/local_velociraptor_driver.py
```diff
@@ -0,0 +1,222 @@
...
+            )
+        return self._cached_query(query)
+
+    @lru_cache(maxsize=256)
+    def _cached_query(self, query: str) -> pd.DataFrame:
+        iter_data_files = (
+            tqdm(self.data_files[query]) if self._progress else self.data_files[query]
+        )
+        dfs = [pd.read_json(file, lines=True) for file in iter_data_files]
+        query_df = pd.concat(dfs)
```
rcobb-scwx, line 165: Unless there is a reason to preserve the index, I recommend passing `ignore_index=True` to `pd.concat` here. This ensures a unique index per row of concatenated results.

msticpy/data/drivers/local_velociraptor_driver.py
```diff
@@ -0,0 +1,222 @@
...
+        Returns
+        -------
+        Dict[str, Dict]
+            Data schema of current connection.
+
+        """
+        if not self._schema:
+            if not self.data_files:
+                self.connect()
+            # read the first row of each file to get the schema
```
rcobb-scwx, line 110: I am not familiar with velocirator, but I assuming that these JSON blobs are not dynamic. Otherwise the schema might miss columns that are not present in the first row of each file.

msticpy/data/drivers/local_velociraptor_driver.py
```diff
@@ -0,0 +1,222 @@
...
+                self._schema[table] = {
+                    col: dtype.name for col, dtype in sample_df.dtypes.to_dict().items()
+                }
+            logger.info("Reading schema for %d tables", len(self.data_files))
+
+        return self._schema
+
+    def query(
+        self, query: str, query_source: Optional[QuerySource] = None, **kwargs
+    ) -> Union[pd.DataFrame, Any]:
```
rcobb-scwx, line 129: Should the return type annotation be `Any` or `results.ResultSet`?

msticpy/data/drivers/local_velociraptor_driver.py
```diff
@@ -0,0 +1,222 @@
...
+            return [
+                {
+                    "name": query,
+                    "query": query,
+                    "query_paths": "velociraptor",
+                    "description": f"Velociraptor {query} table.",
+                }
+                for query in self.data_files
+            ]
+        return []
```
rcobb-scwx, line 203: Since `self.connect` should be assigning `self.data_files`, if `self.data_files` is none then it should throw an exception (or at least generate an error message, rather than fail silently).

tests/data/drivers/test_azure_kusto_driver.py
```diff
@@ -204,17 +204,16 @@ class ConnectTest(NamedTuple):
         init_args={"connection_str": "https://test1.kusto.windows.net"},
         connect_args={"connection_str": "https://test.kusto.windows.net"},
         tests=[
-            lambda driver: driver.client._kusto_cluster
-            == "https://test.kusto.windows.net"
+            lambda driver: driver.client._kusto_cluster.startswith(
```
rcobb-scwx, line 207: Is this meant to be included in this PR? I thought at first that it was just the code formatter, but it looks semantically different from the previous version. 

[Edit: "this" meaning the kusto testing related changes]

tests/testdata/velociraptor/Windows.Forensics.ProcessInfo.json
```diff
@@ -0,0 +1,10 @@
...
...s (x86)\\Common Files","CommonProgramW6432":"C:\\Program Files\\Common Files","COMPUTERNAME":"WKST01","ComSpec":"C:\\Windows\\system32\\cmd.exe","NUMBER_OF_PROCESSORS":"2","OS":"Windows_NT","Path":"C:\\Windows\\system32;C:\\Windows;C:\\Windows\\System32\\Wbem;C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\;C:\\Program Files\\Amazon\\cfn-bootstrap\\;C:\\ProgramData\\chocolatey\\bin;;C:\\Program Files\\Microsoft VS Code\\bin","PATHEXT":".COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC","PROCESSOR_ARCHITECTURE":"AMD64","PROCESSOR_IDENTIFIER":"Intel64 Family 6 Model 85 Stepping 7, GenuineIntel","PROCESSOR_LEVEL":"6","PROCESSOR_REVISION":"5507","ProgramData":"C:\\ProgramData","ProgramFiles":"C:\\Program Files","ProgramFiles(x86)":"C:\\Program Files (x86)","ProgramW6432":"C:\\Program Files","PSModulePath":"%ProgramFiles%\\WindowsPowerShell\\Modules;C:\\Windows\\system32\\WindowsPowerShell\\v1.0\\Modules;C:\\Program Files (x86)\\AWS Tools\\PowerShell\\","PUBLIC":"C:\\Users\\Public","SystemDrive":"C:","SystemRoot":"C:\\Windows","TEMP":"C:\\Windows\\TEMP","TMP":"C:\\Windows\\TEMP","USERNAME":"SYSTEM","USERPROFILE":"C:\\Windows\\system32\\config\\systemprofile","windir":"C:\\Windows"}}
```
rcobb-scwx, line 6: Going back to my earlier comment about how we identify the right schema for each log file, it looks like several columns would be missed here if we only looked at the columns in the first row.

msticpy/data/drivers/local_velociraptor_driver.py
```diff
@@ -0,0 +1,222 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# pylint: disable=too-many-instance-attributes
+@export
+class VelociraptorLogDriver(DriverBase):
+    """OSQueryLogDriver class to execute kql queries."""
+
+    OS_QUERY_DATEIME_COLS = {
```
petebryan, line 33: Given the name are these still valid for Velociraptor?

msticpy/data/drivers/local_velociraptor_driver.py
```diff
@@ -0,0 +1,222 @@
...
+        "unixTime",
+        "columns_time",
+        "columns_atime",
+        "columns_ctime",
+        "columns_mtime",
+    }
+
+    def __init__(self, connection_str: Optional[str] = None, **kwargs):
+        """
+        Instantiate OSQueryLogDriver and optionally connect.
```
petebryan, line 43: Some old/re-used docstrings here

petebryan (approved):

docs/source/data_acquisition/DataProv-OSQuery.rst
juju4, line 1: Mix with osquery provider?

docs/source/data_acquisition/DataProv-Velociraptor.rst
```diff
@@ -0,0 +1,157 @@
+The Velociraptor provider
+=========================
+
+:py:mod:`Velociraptor driver documentation<msticpy.data.drivers.local_velociraptor_driver>`
+
+The ``Velociraptor`` data provider can read Velociraptor log files
```
juju4, line 6: Let's say offline collection log files as per https://docs.velociraptor.app/docs/offline_triage/#offline-collections.
It's a different case if collection logs are sent to a SIEM (ex: https://docs.velociraptor.app/artifact_references/pages/elastic.flows.upload/, https://docs.velociraptor.app/artifact_references/pages/splunk.flows.upload/) which is likely the next step.

msticpy/data/core/query_defns.py
```diff
@@ -106,8 +107,11 @@ class DataEnvironment(Enum):
     Cybereason = 12
     Elastic = 14
     OSQueryLogs = 15
+    OSQuery = 15
     MSSentinel_New = 16
     Kusto_New = 17
+    VelociraptorLogs = 18
+    Velociraptor = 18
```
juju4, line 114: Mix with other PR?
Why two Velociraptor lines with same id?

msticpy/data/drivers/local_velociraptor_driver.py
juju4: both comment and OS_QUERY_DATEIME_COLS variable don't apply to velociraptor so likely not.

msticpy/data/drivers/local_velociraptor_driver.py
juju4: They are or it depends.
Like osquery, the output depends on configured query which is using VQL, a SQL-like format and is customizable by user.
If staying with default artifacts or inside a given collection and artifact, mostly static but not escaping that it can be changed and some edge case like below example logs, likely because of nested data.

See also https://docs.velociraptor.app/docs/vql/artifacts/

docs/source/data_acquisition/DataProv-Velociraptor.rst
```diff
@@ -0,0 +1,157 @@
...
+    {"Pid":540,"Ppid":416,"Name":"services.exe","Path":"C:\\Windows\\System32\\services.exe","CommandLine":"C:\\Windows\\system32\\services.exe","Hash":{"MD5":"fefc26105685c70d7260170489b5b520","SHA1":"d9b2cb9bf9d4789636b5fcdef0fdbb9d8bc0fb52","SHA256":"930f44f9a599937bdb23cf0c7ea4d158991b837d2a0975c15686cdd4198808e8"},"Username":"NT AUTHORITY\\SYSTEM","Authenticode":{"Filename":"C:\\Windows\\System32\\services.exe","ProgramName":"Microsoft Windows","PublisherLink":null,"MoreInfoLink":"http://www.microsoft.com/windows","SerialNumber":"33000002a5e1a081b7c895c0ed0000000002a5","IssuerName":"C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Windows Production PCA 2011","SubjectName":"C=US, ST=Washington, L=Redmond, O=Microsoft Corporation, CN=Microsoft Windows Publisher","Timestamp":null,"Trusted":"trusted","_ExtraInfo":null},"Family":"IPv4","Type":"TCP","Status":"LISTEN","Laddr.IP":"0.0.0.0","Laddr.Port":49728,"Raddr.IP":"0.0.0.0","Raddr.Port":0,"Timestamp":"2022-02-12T19:35:57Z"}
+
+
+The columns in each JSON will be used to create the pandas DataFrame columns.
+
```
juju4, line 64: Let's add a thanks to BlueTeamVillage for the example logs dataset from
https://github.com/blueteamvillage/Project-Obsidian-DC30
https://media.blueteamvillage.org/DC30/Obsidian/

juju4 (commented):
Great work @ianhelle and thanks for the multiple reviews.

tests/testdata/velociraptor/Windows.Forensics.ProcessInfo.json
ianhelle: Not sure that they are. The big thing at the end is the contents of the current environment (although it is itself a json dictionary - not sure I want to cover all eventualities of what might be in a nested field. Most data providers don't populate the schema so I've always considered it a bit of nice addition rather than core functionality.

tests/data/drivers/test_azure_kusto_driver.py
ianhelle: Yep - definite typo

msticpy/data/drivers/local_velociraptor_driver.py
ianhelle: The base class annotations are pd.DataFrame, Any. I sometimes violate this in subclasses where there is something more specific that's reasonable but it isn't ideal.

msticpy/data/drivers/local_velociraptor_driver.py
ianhelle: I'm not sure how expensive we want to make the schema property. Looking at the first record is cheap and cheerful. Even if we did the first 50 or 1000 records, we'd have no guarantee that an extra field was not in record 1001.

msticpy/data/drivers/local_velociraptor_driver.py
ianhelle: Good catch

tests/data/drivers/test_azure_kusto_driver.py
ianhelle: Actually, it isn't a typo. This test started to break on this PR because of a change in the Kusto SDK. This test is checking the value for the URL in the SDK client. This used to be just the cluster URL but became URL+some path or query params.
However, I did find a bug related to this so thanks!!

PR #682, opened 2023-07-18
Added delete_watchlist_item method

mbabinski:
Added a delete_watchlist_item method to sentinel_watchlists.py to enable users to better manage watchlist content with msticpy. For instance, a user may wish to delete watchlist items and then refresh the data in a watchlist. I used the [Sentinel REST API documentation](https://learn.microsoft.com/en-us/rest/api/securityinsights/stable/watchlist-items/delete?tabs=HTTP) for reference.


mbabinski:

ianhelle (approved):

PR #687, opened 2023-07-27
Add AbuseIPDB TIProvider

rrevuelta:
Added AbuseIPDB as a new TI provider.


rrevuelta:
Hi!

In advance, pytest is going to raise an error in "**test_ti_provider.py**" when it checks the results of the search of the IP addresses, the malicious ones (17) and the benign ones (3). The issue is that AbuseIPDB also returns information for benign IPs, with a benign score, and thats why the length of results_df(results_df["Results]) its 20 and not 17.

image

rrevuelta:

ianhelle:
> Hi!
> 
> In advance, pytest is going to raise an error in "**test_ti_provider.py**" when it checks the results of the search of the IP addresses, the malicious ones (17) and the benign ones (3). The issue is that AbuseIPDB also returns information for benign IPs, with a benign score, and thats why the length of results_df(results_df["Results]) its 20 and not 17.
> 
> image
> 
> Thanks in advance!

Thanks @rrevuelta - I'll fix the test so that you can set provider-specific expected results instead of hard-coding.
Looks like a great data source.

ianhelle (approved):

PR #692, opened 2023-08-02
Fix splunk uploader create index option

Tatsuya-hasegawa:
The option name on the document was different from the source code.
I modified it to the name as "create_index" of source code.
It will avoid the error of misuse when user set "create_idx" option name.
Best regards,


Tatsuya-hasegawa:
image
Error example when create_idx was used.

ianhelle (approved):

PR #704, opened 2023-09-01
Update methodology of getting endpoints for cloud environment

ccianelli22:
Will now use arm to get endpoints for each environment
Allow user to get endpoints by listing the arm url, useful if cloud is not global, cn, or usgov 
Update documentation
Deprecate de/Germany since the cloud is also deprecated
Create offline endpoints for global, cn and usgov for mocking 
Remove hardcoded urls that may touch only global endpoints


ianhelle (changes requested):
Awesome - and super thorough work Chris. Thanks so much.
Main thing I noticed is use of requests vs. httpx. You can almost certainly just search for httpx.get in the code and copy/paste an example (along with the imports from common)

ianhelle:
I suspect some of the changes will trigger linter/code format warnings but these should be straightforward to fix

msticpy/auth/cloud_mappings.py
```diff
@@ -5,25 +5,25 @@
 # --------------------------------------------------------------------------
 """Azure Cloud Mappings."""
 import contextlib
+import requests
```
ianhelle, line 8: We switched to using httpx in place of requests but it's almost 100% drop in replacement.
Also have a couple of support utils for timeouts,  user agent string. (actually that reminds me that I need to implement proxy support across these calls.

ianhelle (approved):

PR #709, opened 2023-09-05
Update mde_proc_pub.pkl

FlorianBracq:
With the latest version of pandas (2.1.0), some pickle files could not be parsed with the function `read_pickle`:
* tests/testdata/mde_proc_pub.pkl
* tests/testdata/mordor/mitre_tech_cache.pkl
* tests/testdata/mordor/mitre_tact_cache.pkl

The error triggered in _tests/transform/test_process_tree_utils.py_ and led to all the unit tests in this module to fail.

I'm not sure exactly how those files were generated, but to fix the issue I applied the following steps:
* downgrade to pandas 2.0.3
* Put a breakpoint on where the read_pickle function was called
* Run the test test_process_tree_utils in debug mode
* At the breakpoint, read the pickle file and through the debug console export it with the command:
`df.to_pickle(<filename>)`

Then I upgraded back to 2.1.0 and the file appears to be properly parsed now.


ianhelle (approved):

ianhelle:
Awesome Florian - thanks for fixing. Not sure of the reason for this but presumably something to do with a subtle change in the pandas/pickle object model. in v2.1x. I think this version doesn't support Py3.8, which is why this build continued to work.

PR #758, opened 2024-03-19
Updated conda reqs files for new packages

ianhelle:
Updated files in conda folder to align to new package versions and availability of packages from Microsoft conda repository

### DIFF
--- a/conda/conda-reqs-pip.txt
+++ b/conda/conda-reqs-pip.txt
@@ -1,7 +1,8 @@
-azure-mgmt-monitor>=2.0.0
 azure-mgmt-resourcegraph>=8.0.0
-azure-mgmt-subscription>=3.0.0
-KqlmagicCustom[jupyter-basic,auth_code_clipboard]>=0.1.114.post22
+azure-monitor-query>=1.0.0, <=2.0.0
+# KqlmagicCustom[jupyter-basic,auth_code_clipboard]>=0.1.114.post22
 mo-sql-parsing>=8, <9.0.0
+nest_asyncio>=1.4.0
 passivetotal>=2.5.3
+sumologic-sdk>=0.1.11
 splunk-sdk>=1.6.0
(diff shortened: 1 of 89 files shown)


conda/conda-reqs.txt
```diff
@@ -7,10 +7,11 @@ azure-keyvault-secrets>=4.0.0
 azure-kusto-data>=4.0.0, <=5.0.0
 azure-mgmt-compute>=4.6.2
 azure-mgmt-keyvault>=2.0.0
+azure-mgmt-monitor
```
petebryan, line 10: Do we not want to use the version controls we had for pip here as well? I know azure package versioning has been a compatibility issue in the past.

conda/conda-reqs-pip.txt
```diff
@@ -1,7 +1,8 @@
-azure-mgmt-monitor>=2.0.0
 azure-mgmt-resourcegraph>=8.0.0
-azure-mgmt-subscription>=3.0.0
-KqlmagicCustom[jupyter-basic,auth_code_clipboard]>=0.1.114.post22
+azure-monitor-query>=1.0.0, <=2.0.0
+# KqlmagicCustom[jupyter-basic,auth_code_clipboard]>=0.1.114.post22
```
petebryan, line 3: Whats the reason for commenting out this?

petebryan (approved):

conda/conda-reqs-pip.txt
ianhelle: I'm hoping to remove dependency on Kqlmagic. It's still available as an option but unlike the other optional items in here it has a lot of dependencies. People will get a warning if they try to use one of the deprecated components that still use it and should have enough info to get it installed.

conda/conda-reqs.txt
ianhelle: Unfortunately, when I use standard versioning numbering it fails - this package in the Conda repo uses a date-based version string. Fortunately, conda handles versioning compatibility issues better than pip so it seems to resolve itself.

PR #771, opened 2024-04-16
WSConfig updates

ianhelle:
Main fix is to allow additional connection parameters to be used with MSSentinel QueryProvider.
- Fixing wsconfig
  - updating a lot of legacy code
  - allow link to corresponding settings to persist in WorkstationConfig instance
  - allow it to use Args subkey

MSTICPyConfig
- Can now use MSSentinel instead of AzureSentinel in msticpyconfig.yaml - the two should be interchangeable but it will migrate to the former by default.

azure_monitor_driver/MS Sentinel:
- Can specify additional arguments to pass to connect() function by adding an Args subkey
- This follows the pattern of TI and other providers and supports storage of values in key vault or environment variables
- You can also pass other authentication parameters (such as client_id, client_secret) in the connect method and these will be passed to the underlying az_connect function.

Fix for vtlookup3.py
- Fixed problematic way of using nestasyncio - this was causing failures when run from a langchain agent.
Fix for lookup/tilookup
- If the progress parameter was not passed it would still try to cancel a non-existent progress task and cause an exception.

### DIFF
--- a/msticpy/common/pkg_config.py
+++ b/msticpy/common/pkg_config.py
@@ -409,7 +409,7 @@ def validate_config(mp_config: SettingsDict = None, config_file: str = None):
 
     Parameters
     ----------
-    mp_config : Dict[str, Any], optional
+    mp_config : SettingsDict, optional
         The settings dictionary, by default it will
         check the currently loaded settings.
     config_file : str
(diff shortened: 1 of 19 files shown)


msticpy/common/pkg_config.py
```diff
@@ -390,7 +403,7 @@ def _valid_timeout(timeout_val) -> Union[float, None]:
 refresh_config()
 
 
-def validate_config(mp_config: Dict[str, Any] = None, config_file: str = None):
+def validate_config(mp_config: SettingsDict = None, config_file: str = None):
```
petebryan, line 406: Looks like this needs a doc string update.

petebryan (approved):

PR #777, opened 2024-05-12
Add extra tests and fixes to QueryProvider, DriverBase and (as)sync query handling

FlorianBracq:
Update code to ensure consistency and fix bugs for:
* data_providers
* query_provider_connections_mixin
* driver_base

Those changes are the first steps to some rewrites of the way split queries + additional connections are handled to simplify query caching (#756) and error handling

Test coverage have been increased to 100% (according to pytest) to ensure nothing will break for future updates.

### DIFF
--- a/msticpy/data/core/data_providers.py
+++ b/msticpy/data/core/data_providers.py
@@ -213,42 +213,6 @@ class QueryProvider(QueryProviderConnectionsMixin, QueryProviderUtilsMixin):
         logger.info("Adding query pivot functions")
         self._add_pivots(lambda: self._query_time.timespan)
 
-    def exec_query(self, query: str, **kwargs) -> Union[pd.DataFrame, Any]:
-        """
-        Execute simple query string.
-
-        Parameters
-        ----------
-        query : str
-            [description]
-        use_connections : Union[str, List[str]]
-
-        Other Parameters
-        ----------------
-        query_options : Dict[str, Any]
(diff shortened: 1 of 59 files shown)


msticpy/data/core/data_providers.py
```diff
@@ -212,42 +213,6 @@ def connect(self, connection_str: Optional[str] = None, **kwargs):
         logger.info("Adding query pivot functions")
         self._add_pivots(lambda: self._query_time.timespan)
 
-    def exec_query(self, query: str, **kwargs) -> Union[pd.DataFrame, Any]:
```
ianhelle, line 215: What's the reason for moving this to the mixin class?
No objections, just curious.

ianhelle (approved):
This is v cool indeed. Thx so much Florian.
Especially impressed with the detailed testing.

msticpy/data/core/data_providers.py
FlorianBracq: I moved it here because:
* As the initial empty implementation from *QueryProviderProtocol* of `exec_query` could not be tested, I removed it
* In `_create_split_query_tasks` from *QueryProviderConnectionsMixin*, `exec_query`is invoked thus somethng has to be defined. Between creating an abstract method and using the "real" implementation, I picked the real implementation

I don't mind changing it to something else if needed!

PR #785, opened 2024-06-27
Fixing np.NaN error and build warnings

ianhelle:
Fixing pandas deprecation warnings in vtfile_behaviour and process_tree_utils.
The rest is mostly suppressing expected warnings in unit tests

### DIFF
--- a/tests/context/test_vtlookupv3.py
+++ b/tests/context/test_vtlookupv3.py
@@ -46,7 +46,6 @@ def create_vt_client(vt_lib) -> VTLookupV3:
     """Test simple lookup of IoC."""
     vt_lib.Client = VTClient
     vt_lib.APIError = VTAPIError
-
     return VTLookupV3()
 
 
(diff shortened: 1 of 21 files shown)


tests/context/test_vtlookupv3.py
```diff
@@ -161,22 +160,6 @@ def close(self):
         return None
 
 
-def test_init_vt_lookup_class():
```
ryan-detect-dot-dev, line 164: Was this deleted on purpose?

ryan-detect-dot-dev (approved):
LGTM

tests/context/test_vtlookupv3.py
ianhelle: Yes it was. It was a trivial test and needed some nasty fix to instantiate VTLookup in an async method. 
I didn't think the effort to change the test was worth the value that the test brought.

FlorianBracq:
That looks good to me too!

PR #791, opened 2024-07-05
Fixing MSSentinel to obey parameters

ianhelle:
- multiple updates to sentinel_core and sentinel_utils - to correctly adhere to parameters and not default to using config workspaces, when overridden
- removed remaining references to msrestazure
- using azure.mgmt.core.tools to parse and validate resource IDs
- deprecating use of sub_id to set default subscription and set default workspace in MicrosoftSentinel
- Sentinel Incidents get_incident now returns full alert details when alerts=true


msticpy/context/azure/sentinel_core.py
```diff
@@ -35,6 +39,59 @@
...
+    "ws_name": _WS_NAME,
+    "workspace": _WS_NAME,
+    "res_id": _RES_ID,
+}
+_CORE_WS_PARAMETERS = [_SUB_ID, _RES_GRP, _WS_NAME]
+_WS_PARAMETERS = _CORE_WS_PARAMETERS + [_RES_ID]
+
+
+def _map_legacy_param_names(**kwargs) -> Dict[str, Any]:
+    """Map legacy parameter names to current names."""
```
ryan-detect-dot-dev, line 92: It might be worth expanding on this docstring because of the heavy lifting that this function is going:

- Historically, msticpy used abbreviated versions of the kwarg parameters needed to connect to a Sentinel instance
- These names have since changed, but we want to accept the old kwargs and map them behind the scenes to their new names
- This function will search through the passed-in kwargs to find these outdated kwarg names and replace them with their modern equivalents. If not found, then it simply passes the kwarg along as-is.

msticpy/context/azure/sentinel_core.py
```diff
@@ -54,97 +111,102 @@ class MicrosoftSentinel(
...
+        if resource_id:
+            init_kwargs[_RES_ID] = resource_id
+        if subscription_id:
+            init_kwargs[_SUB_ID] = subscription_id
+        if resource_group:
+            init_kwargs[_RES_GRP] = resource_group
+        if workspace_name:
+            init_kwargs[_WS_NAME] = workspace_name
+
+        self._default_settings = self._set_ws_defaults(_get_ws_defaults, **init_kwargs)
```
ryan-detect-dot-dev, line 174: This is probably worth a type annotation:

`self._default_settings: Callable[..., SentinelInstanceDetails] = self._set_ws_defaults(_get_ws_defaults, **init_kwargs)`

msticpy/context/azure/sentinel_core.py
```diff
@@ -54,97 +111,102 @@ class MicrosoftSentinel(
...
-            self._resource_id = self._build_sent_res_id(
-                sub_id=self.default_subscription,
-                res_grp=self._default_resource_group,
-                ws_name=workspace_name,
-            )
-            logger.info("Resource ID set to %s", self._resource_id)
-            self._default_workspace = workspace_name
-            self.url = self._build_sent_paths(
-                self._resource_id, self.base_url  # type: ignore
+        if not any(ws_param in init_kwargs for ws_param in _WS_PARAMETERS):
```
ryan-detect-dot-dev, line 192: I feel like this block should be in the `_get_ws_defaults` or `_set_ws_defaults` functions somewhere.

msticpy/context/azure/sentinel_core.py
```diff
@@ -199,57 +283,63 @@ def connect(
...
-            workspace_name=self.workspace_config.get(WorkspaceConfig.CONF_WS_NAME_KEY),
-        )
-
     def _create_api_paths_for_workspace(
         self,
         az_resource_id: Optional[str] = None,
-        subscription_id: Optional[str] = None,
-        resource_group: Optional[str] = None,
-        workspace_name: Optional[str] = None,
+        subscription_id: str = "",
```
ryan-detect-dot-dev, line 305: I think these params and the arguments to `self._build_sent_res_id` should just take an instance of `sentinel_instance_details: SentinelInstanceDetails`?

msticpy/context/azure/sentinel_core.py
```diff
@@ -199,57 +283,63 @@ def connect(
...
     ):
         """Save configuration and build API URLs for workspace."""
-        if workspace_name:
-            self.workspace_config = WorkspaceConfig(workspace=workspace_name)
         az_resource_id = (
             az_resource_id
-            or self._resource_id
             or self._build_sent_res_id(
-                subscription_id, resource_group, workspace_name  # type: ignore
+                subscription_id or self.default_subscription_id or "",
```
ryan-detect-dot-dev, line 313: Shouldn't this just be `az_resource_id or sentinel_instance_details.resource_id`?

msticpy/context/azure/sentinel_core.py
```diff
@@ -259,87 +349,114 @@ def _create_api_paths_for_workspace(
...
 
+        Sets the default workspace to use for subsequent queries. If the `workspace`
+        parameter is specified, the workspace settings will be read from the
+        msticpyconfig file. If the `subscription_id` parameter is specified, it will return
+        the workspace from the subscription, if only one workspace is found.
+        If neither are specified but a workspace resource ID has been set during
+        initialization (the `resource_id` parameter), it will use the workspace details
+        from that resource ID.
+        Otherwise it will set the workspace to be the default workspace in the
+        configuration file.
```
ryan-detect-dot-dev, line 374: I need a flow chart :D

msticpy/context/azure/sentinel_core.py
```diff
@@ -35,6 +39,59 @@
...
+_RES_GRP = "resource_group"
+_WS_NAME = "workspace_name"
+_RES_ID = "resource_id"
+
+
+@dataclass
+class SentinelInstanceDetails:
+    """Dataclass for Sentinel workspace details."""
+
+    subscription_id: Optional[str]
```
ryan-detect-dot-dev, line 52: I am not a fan of making these optional. The point of this pattern is that the `SentinelInstanceDetails` can only instantiate when it has the necessary arguments. If any of these are not passed, throwing an exception is the right thing to do.

msticpy/context/azure/sentinel_core.py
```diff
@@ -35,6 +39,59 @@
...
+    resource_group: Optional[str]
+    workspace_name: Optional[str]
+
+    @property
+    def resource_id(self) -> Optional[str]:
+        """Return the resource ID for the workspace."""
+        if not all([self.subscription_id, self.resource_group, self.workspace_name]):
```
ryan-detect-dot-dev, line 59: Unnecessary if the class needs all the args to instantiate.

msticpy/context/azure/sentinel_core.py
```diff
@@ -35,6 +39,59 @@
...
+            return None
+        return build_sentinel_resource_id(
+            self.subscription_id, self.resource_group, self.workspace_name  # type: ignore
+        )
+
+
+def _get_ws_defaults(
+    *,
+    subscription_id: Optional[str] = None,
```
ryan-detect-dot-dev, line 68: Again, I don't think these should be Optional?

When this is called, all 3 arguments _must_ be present. Each time we partial this function, we can potentially add one or more of them, but that doesn't make them optional.

ryan-detect-dot-dev (commented):
Looks much better than before. Let's chat about this PR next we meet.

msticpy/context/azure/sentinel_core.py
ryan-detect-dot-dev: Amendment: they might be Optional here if we also checked for "none of the args are passed, so I will do the default workspace lookup instead" before passing it into constructor for the SentinelInstanceDetails

msticpy/context/azure/sentinel_core.py
```diff
@@ -100,15 +111,17 @@ def __init__(
...
 
         workspace = kwargs.get("workspace", ws_name)
-        self._default_workspace: Optional[str] = workspace
+        self._default_workspace_name: Optional[str] = workspace
         self.workspace_config = WorkspaceConfig(workspace)
 
         logger.info("Initializing Microsoft Sentinel connector")
         logger.info(
-            "Params: Cloud=%s; ResourceId=%s; Workspace=%s",
+            "Params: Cloud=%s; ResourceId=%s; Workspace=%s, Subscription=%s, ResourceGroup=%s",
```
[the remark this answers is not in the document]
ianhelle: Actually, logger docs say *not* to use f-strings. The reason being (I think) is because f-strings are always evaluated. The logger substitutions are only evaluated if the logger level is at the level to trigger the call. It essentially makes uncalled logging calls cost-free.

msticpy/context/azure/sentinel_core.py
ianhelle: agreed.

msticpy/context/azure/sentinel_core.py
ianhelle: I've changed the logic to do this in _set_ws_defaults

msticpy/context/azure/sentinel_core.py
ianhelle: I think this is now moot since the code has disappeared

ryan-detect-dot-dev (commented):
LGTM! Much cleaner!

ryan-detect-dot-dev (approved):
LGTM

PR #794, opened 2024-07-19
Update TILookup and ContextLookup

FlorianBracq:
Update lookups to expand kwargs to ensure keywod arguments are named and visible.
Add typing for pa to most of the functions to ensure consistency and proper mypy support.
Apply ruff recommendations:
  * Typing with list/tuple/dict/set instead of List/Tuple/Dict/Set
  * Add ClassVar types
  * Replace Optional and Union by new "pipe notation"
  * Add timezone to datetime objects


msticpy/context/vtlookupv3/vtlookup.py
```diff
@@ -190,7 +197,7 @@ def lookup_iocs(
         src_col: str = "Observable",
         type_col: str = "IoCType",
         src_index_col: str = "SourceIndex",
-        **kwargs,
+        **kwargs: str,
```
ianhelle, line 200: kwargs shouldn't be a str right?
Maybe you could do a quick sweep for kwargs being mis-typed.

ianhelle:
Wow Florian. You've gone above and beyond the call of duty here. Amazing work!
I'm assuming/hoping that you had some tool assistance with this.

ianhelle (changes requested):
Just the kwargs: str type annotation

conda/conda-reqs-pip.txt
```diff
@@ -7,3 +7,4 @@ passivetotal>=2.5.3
 sumologic-sdk>=0.1.11
 splunk-sdk>=1.6.0,!=2.0.0
 packaging>=24.0
+requests>=2.31.0
```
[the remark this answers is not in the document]
FlorianBracq: It's related to RiskIQ/PassiveTotal, yes.
As we are exposing the property "_httpx_client" in msticpy/context/tiproviders/riskiq.py lines 153 to 162 and those are requests.Session objects, not httpx.Client, there is still a dependency to requests although we migrated to httpx.

msticpy/context/vtlookupv3/vtlookup.py
FlorianBracq: I'm also planning to remove (most of) kwargs in a future PR, but in some cases such as this one, I do not know the list of keywords we could have, nor if it's a finite list

ianhelle: It's def not treated as a string.  line 256 iterates through the kwargs dict.
Having said that, this is a very old bit of code which I'm not sure anyone uses (vs the TIProvider implementation, so I'd be open to jettisoning...maybe even deprecating the class altogether.

I'm a bit concerned about the idea of removing kwargs everywhere. I think it's OK where they are only used one level deep but sometimes (like data providers), stuff that is unknown to the outer methods gets consumed by something deeper down the stack.

FlorianBracq: Sorry, it seems there is a small misunderstanding there. The [official doc](https://typing.readthedocs.io/en/latest/spec/callables.html#annotating-args-and-kwargs) indicates that type for **kwargs indicates the type of the values from the keyword arguments, not the "real" type of **kwargs which would indeed be a dictionary-like object. Here, **kwargs; int should be understood as dict[str, int].

There are deeper explanations for TypedDict and Unpack that I am not really familiar with yet in that doc that should be leveraged when kwargs cannot be removed, as for dataproviders indeed. I feel it would make the code a bit longer but the extra added value for clarity is worth it. Happy to tone it down if you believe I'm going too far down that rabbit hole.

ianhelle: Oh! I didn't know that **kwargs: str meant dict where values are of type str!
Cool. Ok that looks good then.

I'm happy with the replacing kwargs idea where it is pretty straighforward. It def gets trickier when you have a facade class like QueryProvider. 
Often I've used kwargs where I think we might need to extend functionality without messing with the signature. I guess as long as the named args remain optional though, it doesn't cause any backward compat issues (and makes the signatures and docs more understandable.
Happy to discuss on a case-by-case basis.

ianhelle (approved):

PR #799, opened 2024-09-18
Fix for bug in azure_core_auth that fails authentication

ianhelle:
Something happened during Florian's renaming to change the logic such that it always failed.
Adding test case for this function

### DIFF
--- a/msticpy/auth/azure_auth_core.py
+++ b/msticpy/auth/azure_auth_core.py
@@ -351,37 +351,28 @@ def _az_connect_core(
     azure_identity_logger.handlers = [handler]
 
     if not credential:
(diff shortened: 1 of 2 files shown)


FlorianBracq:
Sorry, it seems my updaet was borked :(

I have not completely wrapped my head around the proper types for AzCredential.
Should legacy be:
* a "TokenCredential" to be compatible with the calls in method __legacy_auth_ from msticpy/context/azure/azure_data.py (l1068 and 1078)
* or a CredentialWrapper to be aligned with how it is used in that file?

Apart from that, it looks ok for me!

FlorianBracq (approved):

PR #809, opened 2024-12-06
Fixing azure_monitor_driver for deprecated httpx API

ianhelle:
httpx has deprecated/removed `proxies` parameter in API call.
Replacing with `mounts` in httpx.Client
Also fixed a doc bug in process_tree_utils and initializer in host_logon_session.


ianhelle (commented):
note - respx calls in unit tests are failing due to change in new release of httpx.
If respx update release takes a while, I will fix here

FlorianBracq:
> note - respx calls in unit tests are failing due to change in new release of httpx. If respx update release takes a while, I will fix here

I have noticed typing issues that were failing the pipeline too, those should be fixed now!

FlorianBracq (approved):

PR #817, opened 2025-01-14
add Cyberint TI provider

vx3r:
This TI provider is using the [Cyberint](https://cyberint.com/) API for IoC lookup.


ianhelle (approved):
I wasn't really able to verify this against the API but it looks reasonable.
The only thing that I was unsure about was the {Instance} variable in base URL - presumably this is, like ServiceNow, where you have an instance of the API per customer or something like that?
You might want to publicize its existence by addin something to the docs\source\data_acquisition\TIProviders.rst doc - but it's up to you.

Many thanks for the submission!!!

PR #826, opened 2025-02-06
Updating bokeh code to support 3.4.0+

ianhelle:
Main fixes were changes to Circle and plot.circle to use "radius" vs. "size" for the node size.
Also, updated bokeh to >=3.4.0+ in the requirements files.
Moved panel to be a core dependency.


msticpy/vis/entity_graph_tools.py
```diff
@@ -33,6 +35,8 @@
 req_alert_cols = ["DisplayName", "Severity", "AlertType"]
 req_inc_cols = ["id", "name", "properties.severity"]
 
+_BOKEH_VERSION = tuple(int(i) for i in bokeh.__version__.split("."))
```
FlorianBracq, line 38: I would implement this with:
```python
from importlib.metadata import version
from packaging.version import Version, parse

_BOKEH_VERSION: Version = parse(version("bokeh"))
```

msticpy/vis/entity_graph_tools.py
```diff
@@ -508,15 +512,26 @@ def plot_entitygraph(  # pylint: disable=too-many-locals
     graph_renderer = from_networkx(
         entity_graph_for_plotting, nx.spring_layout, scale=scale, center=(0, 0)
     )
+    if _BOKEH_VERSION > (3, 2, 0):
```
FlorianBracq, line 515: If _BOKEH_VERSION is typed as Version, we can use:
```python
_BOKEH_VERSION > Version("3.2.0")
```

FlorianBracq (commented):
I would modify the computation of _BOKEH_VERSION and comparison, but apart from that it looks of for me!

msticpy/vis/entity_graph_tools.py
ianhelle: As always, I can rely on you for a more elegant solution 💯

FlorianBracq (approved):

PR #840, opened 2025-04-01
Compat fixes for Bokeh 3.7

ianhelle:
Also suppressing classes of mypy errors in bokeh modules since it is broadly incompatible with Bokeh.
Fixing flake8 error in common/utilty/types


juju4:
LGTM. Thanks @ianhelle

ashwin-patil (approved):

PR #847, opened 2025-05-13
Prismacloud_aws_detection_enhancement

raj-axe:
+ various additional detections enhancement in the support queries
+ changes to limits in config query  to pull more data

### DIFF
--- a/msticpy/data/drivers/prismacloud_driver.py
+++ b/msticpy/data/drivers/prismacloud_driver.py
@@ -928,7 +928,7 @@ class PrismaCloudDriver(DriverBase):  # pylint: disable=R0902
         max_retries = int(kwargs.get("max_retries", self.max_retries))  # type: ignore[arg-type]
 
         logger.info("Executing configuration search with query: %s", query)
-        payload = {"query": query, "limit": 1000, "withResourceJson": True, **kwargs}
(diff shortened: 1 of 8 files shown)


ianhelle:
Hmm. it occurred to me that the yaml format you are using might cause some issues. There seem to be 2 tests that are failing in Py3.9 and Py3.11 pipelines.
I'm kind of OK to get rid of or loosen the json_schema check. 
You should check the tests/data/queries/test_query_files.py failure though.
If it is not trivial to fix, I suggest that we just bypass the prismacloud queries in the package tests.

In tests/common/test_pkg_config.py, something like adding an exception for prismacloud in the for loop:
```
    if "prismacloud" in str(query_file):
        continue
```
like this

```python
def test_load_default():
    """Test load default settings."""
    settings = pkg_config._settings
    check.is_in("QueryDefinitions", settings)
    check.is_in("Default", settings["QueryDefinitions"])
    check.equal(1, len(settings["QueryDefinitions"]["Default"]))
    for path in settings["QueryDefinitions"]["Default"]:
        check.is_true(type(path), str)
        path = (
            Path(pkg_config.__file__).resolve().parent.parent.joinpath(f"data/{path}")
        )
        check.is_true(path.is_dir())
        for query_file in path.rglob("*.yaml"):
            if "prismacloud" in str(query_file):
                continue
            validate_queries_file_structure(query_file)

```

PR #849, opened 2025-05-18
[nbwidgets] Querytime - Fix python 3.12 deprecation warning+ apply typing

FlorianBracq:
* Update nbwidgets.query_time to set timezone info in datetime objects
* Apply typing to variables and function arguments
* Explode kwargs


FlorianBracq:
It doesn't look like the linting errors are related to my updates, how would you like those to be tackled?

ianhelle (approved):

PR #853, opened 2025-08-08
Updating Python build versions

ianhelle:
Adding Py 3.12 to build (postponing Py 3.13 for the moment).

Primary changes:
- remove dependencies on pkg_resources
- replace datetime.utcnow (and similar) with datetime.now(tz=timezone.utc)
- updating some type annotations and adding from __future__ import annotations.
- removing KeyVault create_vault (old code dependent on ADAL authentication and not really needed in package).


msticpy-docker-test/msticpy-docker-test/Dockerfile
FlorianBracq, line 1: Was this file intended to be uploaded?

FlorianBracq (approved):

PR #856, opened 2025-09-15
Move Defender APIs to OAuth2, fix invalid scopes

ianhelle:


FlorianBracq (approved):
I slightly changed the 'schema' test and applied some minor linting changes.
Also tested the code on my env, and all seems to be working well!

As a side note, some parameters from the __init__ are not documented, I don't know if it's worth addressing in this PR?

PR #880, opened 2026-02-11
Update Cybereason driver to use structural pattern matching

FlorianBracq:
Starting with python 3.10, structural pattern matching was introduced through [PEP 636](https://peps.python.org/pep-0636/) and it helps hanlding (among many things) dictionaries/JSON responses from API.

The goal of this PR is to update the Cybereason driver to leverage this new feature and ensure more corner cases of the Cybereason API are covered.
Changes include:
* Improve support of timeout errors by adding a retry
* improve parsing of (what appears to be a) successful response
* improve handling of HTTP 429 (too many requests)
* Update docstring
* Change exceptions to leverage msticpy's exceptions as much as possible

Another change is in _query_provider_connections_mixin_ to update the logging of exception handling:
* Add more details for a global exception
* Add specific logging for MsticpyDataQueryError .


ianhelle (approved):
Pattern matching! What a delight!

PR #882, opened 2026-03-17
Replace msrestazure by azure-mgmt-core

FlorianBracq:
Package msrestazure has been deprecated for some time and been replaced by azure-mgmt-core.
This "new" package was already present in some requirement files (requirements-all.txt), so it should not be a big change!

I remove (almost) all references I found to this package, except one in _NOTICE.txt_, I'm not sure how you handle those licenses?

https://pypi.org/project/msrestazure/


FlorianBracq:
I was not expecting the CloudError exceptions to be dependent on msrestazure too...
I replaced those with what made sense to me, but that may not be the best solution. Should some of the Sentinel/blob storage content be updated to use the official SDK instead (in a different PR!)?

ianhelle (approved):

PR #883, opened 2026-03-25
Add Sentinel certificate auth support and docs

ianhelle:
Adding support for certificate authentication to sentinel query provider.
Also updating docs for this an Azure authenticaiton with certs.


FlorianBracq (approved):
Looks good!

PR #889, opened 2026-04-27
Ianhelle/tar slip did protection 2026 04 27

ianhelle:
Some defense-in-depth fixes for handling zip/tar files.
For msticpy all of the files used are secured on https endpoints (trusted github repos - Azure-Sentinel and OTRF - and Maxmind) so there is no opportunity for an attacker to inset malicious payloads.
However, the code change to validate target paths for extracted files is minimal so have do that here.

Also updating some pickle files used in tests to updated numpy/pandas version to avoid deprecation warning.


msticpy/common/archive_utils.py
```diff
@@ -0,0 +1,190 @@
...
+
+    """
+    member_path = Path(member_name)
+    if member_path.is_absolute() or member_name.startswith("/"):
+        raise MsticpyUserError(
+            f"Archive member has an absolute path '{member_name}'.",
+            "This may indicate a malicious archive (path traversal attack).",
+            title="Unsafe archive member path",
+        )
+    if ".." in member_path.parts:
```
FlorianBracq, line 56: This seems redundant with the next (is_relative_to) test?

msticpy/data/drivers/sentinel_query_reader.py
```diff
@@ -131,14 +132,15 @@ def get_sentinel_queries_from_github(
...
         # Only extract Detections and Hunting Queries Folder
-        for file in archive.namelist():  # type: ignore
-            if file.startswith(  # type: ignore
-                (
-                    "Azure-Sentinel-master/Detections/",
-                    "Azure-Sentinel-master/Hunting Queries/",
-                )
-            ):
-                archive.extract(file, path=outputdir)  # type: ignore
+        with archive:  # type: ignore
```
FlorianBracq: I'm not sure why a context manager is being used here?

FlorianBracq (changes requested):

FlorianBracq (approved):

PR #923, opened 2026-08-03
Fix scopes for SecurityGraph provider

N4k0r:
Hello,

When trying to use the `SecurityGraph` provider through an application, I am getting the following error:

```
auth_result (dict) = {"error":"invalid_scope","error_description":"AADSTS1002012: The provided value for scope User.Read is not valid. Client credential flows must have a scope value with /.default suffixed to the resource identifier (application ID URI). [...]
```

The `["User.Read"]` scopes in the driver is what is causing the issue.

When `scopes` is overridden to `["https://graph.microsoft.com/.default"]` with the code below, the authentication is successful:

```
security_graph_provider = QueryProvider("SecurityGraph")
query_provider = security_graph_provider._query_provider
query_provider.scopes = ["https://graph.microsoft.com/.default"]
security_graph_provider.connect()
```

### DIFF
--- a/msticpy/data/drivers/security_graph_driver.py
+++ b/msticpy/data/drivers/security_graph_driver.py
@@ -39,8 +39,8 @@ class SecurityGraphDriver(OData):
         """
         super().__init__(**kwargs)
         az_cloud_config = AzureCloudConfig(cloud=kwargs.pop("cloud", None))
-        self.scopes = [f"{self.api_root}.default"]
         self.api_root = az_cloud_config.endpoints.get("microsoftGraphResourceId")
+        self.scopes = [f"{self.api_root}.default"]
         self.req_body = {
             "client_id": None,
             "client_secret": None,  # nosec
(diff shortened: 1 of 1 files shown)


ianhelle (approved):

msticpy/data/drivers/security_graph_driver.py
```diff
@@ -39,7 +39,7 @@ def __init__(self, connection_str: str | None = None, **kwargs):
         """
         super().__init__(**kwargs)
         az_cloud_config = AzureCloudConfig(cloud=kwargs.pop("cloud", None))
-        self.scopes = ["User.Read"]
+        self.scopes = [f"{self.api_root}.default"]
```
ianhelle: This line needs to go after the next one. Otherwise self.api_root is None and scopes would have "None.default" - which doesn't look like it would work.

Thanks for the fix though. Pls re-submit and I'll approve and merge

ianhelle (changes requested):

msticpy/data/drivers/security_graph_driver.py
N4k0r: Yes you're right, sorry about that.

I did a new commit to fix this,

ianhelle (approved):

PR #937, opened 2026-09-13
feat: add query() time_unit option, list_streams(), list_alerts(), docs fix

juju4:


msticpy/data/drivers/openobserve_driver.py
```diff
@@ -204,7 +204,8 @@ def _query(
 
         """
         del query_source
-        self._ensure_connected()
+        if not self._connected:
+            raise self._create_not_connected_err("OpenObserve")
```
ianhelle, line 208: This changes the established not-connected exception title from `OpenObserveDriver` to `OpenObserve`, causing `test_openobserve_query_no_connect` to fail across the CI matrix. Please restore `self._ensure_connected()` (or preserve the existing driver-name behavior) rather than manually constructing a differently named error.

docs/notebooks/Openobserve-DataConnector.ipynb
```diff
@@ -58,6 +58,7 @@
   },
   {
    "cell_type": "markdown",
+   "execution_count": 1,
```
ianhelle, line 61: This adds `execution_count` to a Markdown cell, which is invalid under the notebook schema. `nbformat.validate()` raises `NotebookValidationError: Additional properties are not allowed`. Please remove this property from the Markdown cell.

msticpy/data/drivers/openobserve_driver.py
```diff
@@ -423,3 +429,39 @@ def _get_openobserve_settings(
...
+        None
+
+        Returns
+        -------
+        tuple[pd.DataFrame, Any]
+            A DataFrame (if successful) or
+            the underlying provider result if an error occurs.
+
+        """
+        df_streams = self.service.list_objects2df("streams")
```
ianhelle, line 448: `list_streams()` and `list_alerts()` both dereference `self.service` without verifying that the driver is connected. Before `connect()`, these methods raise `AttributeError` instead of the standard `MsticpyNotConnectedError`. Please call `_ensure_connected()` in both methods and add coverage for their connected and unconnected behavior.

docs/source/data_acquisition/DataProv-OpenObserve.rst
```diff
@@ -0,0 +1,116 @@
...
+.. code:: ipython3
+
+    df_streams = openobserve_prov.list_streams()
+    df_streams[df_streams['stream_type'] == 'logs'][['name']].head()
+
+    query = """SELECT host_name as "host_name",
+       min(_timestamp) as "firstseen",
+       max(_timestamp) as "lastseen",
+       count() as "count"
+       FROM "journald"  GROUP BY host_name
```
ianhelle, line 93: The new examples are not executable: they use `openobserve_prov` although the provider above is named `qry_prov`, this SQL string closes with four quotes, and the following example never closes its triple-quoted string. Please correct both examples so they can be copied and run.

docs/source/data_acquisition/DataProv-OpenObserve.rst
```diff
@@ -0,0 +1,116 @@
...
+    )
+    df.head()
+
+Other OpenObserve Documentation
+-----------------------------
+
+For examples of using the OpenObserve provider, see the sample
+`OpenObserve Notebook <https://github.com/microsoft/msticpy/blob/master/docs/notebooks/OpenObserve-DataConnector.ipynb>`
+
+Built-in :ref:`data_acquisition/DataQueries:Queries for OpenObserve`.
```
ianhelle, line 114: This `:ref:` target does not exist. Sphinx reports `undefined label: data_acquisition/dataqueries:queries for openobserve`. Please link to an existing label, or use `:doc:` if the intent is to link to the DataQueries page.

ianhelle (changes requested):
The feature additions look reasonable, but the PR currently fails tests and introduces invalid notebook/documentation content. Please address the inline findings before merge.

PR #938, opened 2026-09-16
Add ScanMalware threat intelligence provider

jonaslejon:
## What this adds

A `ScanMalware` threat intelligence provider. [ScanMalware.com](https://scanmalware.com) renders a
submitted URL in a sandboxed browser and publishes the resulting verdict, so a lookup answers "has
this been scanned, and what did the scanner conclude".

Supported observables: `dns`, `hostname`, `url`, `ipv4`, `ipv6`. No file hashes.

**The API needs no key.** A key only raises the rate limit (600 req/min anonymous, 3000 with one),
so the config entry has no `Args` block, following GreyNoise's community-API pattern:

```yaml
TIProviders:
  ScanMalware:
    Primary: True
    Provider: "ScanMalware"
```

## Two design choices worth explaining

**`_run_ti_lookup_query` is overridden because a verdict needs two requests.** The scan list for a
host carries no verdict field, and `security_verdict` lives on `/api/v1/result/{scan_id}`, keyed by
an id that has to be obtained first. I checked every single-request alternative
(`/domain/stats/{host}`, SMQL rows, `/domain/{host}`, `/domains/{host}/latest`,
`/domain/verdict/{host}`) and none carries a verdict, so a purely declarative `_QUERIES` provider
could only report a scan count, which is not a threat signal. `parse_results` is left pure. This
follows the existing precedent in `pulsedive`, `riskiq`, `open_page_rank` and `tor_exit_nodes`.

**The `-quick` query subtypes skip the second request**, for bulk runs that only need to know
whether an observable has been seen. This could not be a config `Args` key: `lookup.py` filters
provider kwargs against `_HTTP_PROVIDER_LEGAL_KWARGS` and silently drops anything else, so such a
key would look like it worked and do nothing. Query subtypes are the supported mechanism, as used
by GreyNoise's `ipv4` / `ipv4-quick` / `ipv4-full`. `ipv4-quick` and `ipv6-quick` exist so a mixed
DataFrame of hosts and addresses works with a single `ioc_query_type="quick"`.

## Correctness details

**A host's scan list contains scans of other pages that merely redirected to it, and a redirector
is frequently the newest row.** Taking `scans[0]` would hand an innocent host the verdict of
whatever page last pointed at it. The provider selects on the API's own `matched_on` field instead.
Live example as of today, for `example.com`:

```
guarded        matched_on=['url', 'final_url']   https://example.com
naive rows[0]  matched_on=['final_url']          https://zestylucas.cha-rlo-ttebird-78c.workers.dev/redirect?...
```

Both happen to carry the same risk level today, so that pair is not itself a demonstration of harm;
the unit test pins the harmful case with a fixture that puts a redirector first on purpose.

**URL observables additionally require an exact URL match**, because a verdict on `site.com/a` says
nothing about `site.com/b`.

**Addresses get passive-DNS context only and never a verdict**, since an address is not a thing that
gets scanned and lending it one page's risk level would tar every other site on that host. The
`/ct/ip` response is unbounded, so the domain list in `Details` is capped.

**Severity mapping is lossy and commented as such.** `ResultSeverity` has no `malicious`, so
`malicious` / `critical` / `high` all collapse onto `high`, with the un-collapsed wording preserved
in `Details["Verdict"]`.

## Verification

- `pytest tests/context/test_tiproviders.py` : **40 passed** on unmodified `main`, **41 passed**
  with this change.
- `pytest tests/context/test_scanmalware.py` : **21 passed**. 62 across both suites.
- `ruff check` and `ruff format --check` clean; `pylint` **10.00/10** on the new module.
- `pytest tests/context/` reports 11 collection errors both with and without this change
  (missing `azure.mgmt.network` and a vtlookupv3 dependency, pre-existing and unrelated).
- Verified end to end against the live API, and confirmed in the service's own nginx logs that both
  requests arrive identifying themselves as `scanmalware-msticpy/1.0`.

`CONTRIBUTING.md` asks for 80% coverage on new code. I could not measure coverage locally (every
tracer dies with `numpy: cannot load module more than once per process` in my environment), so I am
not claiming a number; CI will measure it. The 21 tests cover the attribution guard, every risk
level including an unrecognised one, URL exact-match and trailing-slash, `-quick`, addresses, a
failed verdict hop degrading rather than raising, and the User-Agent on both requests.

## A note on the User-Agent

The provider declares its own `User-Agent` header rather than relying on `mp_ua_header()`. That is a
supported opt-out: `http_provider.py` only applies the default `if "User-Agent" not in
req_dict["headers"]`. It is also necessary here, because the second request does not pass through
`_substitute_parms` and would otherwise send httpx's default.

Separately, `mp_ua_header()` returns the key `"UserAgent"` rather than `"User-Agent"`, so msticpy's
own identifier has not actually been reaching services as an HTTP header. That is unrelated to this
provider and I have offered a fix for it in #936; this branch is deliberately scoped to the provider
and does not touch it.

## Disclosure

I work on ScanMalware.com (Triop AB). Happy to adjust anything here, including dropping the provider
from the default config or changing how the two-request behaviour is exposed.


ianhelle (approved):
Nice addition - thanks for the contribution

